PR #15, opened 2015-09-23
Fix compile errors on the example app

### LINKED ISSUE #14: Fix compiler errors on the example-app
Currently the tsc command produces 2 compile time errors related to the Angular 2.0 debug classes that we are importing.

igor-ka:
This will resolve #14


bertrandk:
:shipit:

PR #18, opened 2015-09-24
wip: render component tree view in html like syntaxt

igor-ka:
- add systemjs based module loading
 - add immutable library
 - restructure the code for component inspector
 - create the angular panel for the tree view


bertrandk:
Looks good :shipit:

I'm kind of bummed about CSP restrictions here thought.

igor-ka:
yeah i tried to work around it but didnt focus on this too much.
We might be able to relax some restrictions by modifying some fields in the manifest.json file, but i did not explore it fully.

PR #24, opened 2015-09-29
fix: compile errors related to typings

### LINKED ISSUE #23: Compile error due from typings
Need to install all the typings dependencies from tsd install as opposed to some being linked from the node_modules folder using tsd link. This way the typings can reference each other from a common location.

For example angular2 comes with a typings for es6-promise with angular2 referencing it. Also chrome typings installed from tsd repo have es6-promise as a dependency, typings for which get installed into /typings folder.

During compilation both the es6-promise from node_modules/angular2 and es6-promise from typings folder (referenced by chrome.d.ts) are present on the build path, causing a compile time error of duplicate occurrence of symbol.

igor-ka:
- install all the necessary typings from tsd so as internal typing references work.
 - hardcode immutable in tsd.d.ts (no available via tsd)
 - add angular2 as dependency in package.json (was ommited on last commit)
 - modify .gitigore to exclude tsd.d.ts

This PR will resolve #23

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 app/js
-app/typings
+app/typings/*
+!app/typings/tsd.d.ts
 example-app/js/*
 example-app/node_modules/*
 example-app/typings/*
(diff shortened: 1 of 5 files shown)


vanessayuenn:
Looks good! :shipit:

PR #32, opened 2015-10-19
feat(adapters): Use tagnames when generic comps

bertrandk:
In cases where the component isn't backed by a framework or user-defined
class, we'll use the tagname for display. This prevents showing
undistinguishable "Object" annotations for the component view.

NOTE: In HTML documents, the casing of tag/node names isn't preserved
and is uppercased instead. If this isn't desirable, the front-end should
format the display as it sees fit.


igor-ka:
love it :ship:

PR #33, opened 2015-10-19
fix(adapters/controllers): track changes to views

bertrandk:
We're adding a new adapter event here: 'clear'. For the time being, rather
than attempt to keep track of internal views at the DOM layer, we'll be
clearing and rebuilding the tree model whenever a change is observed.

This reduces a lot of the complexity involved in reconciling the view
listener's world and that of the resultant DOM.

I've also done some code cleanup.


igor-ka:
@bertrandk looks good :ship:

Issue #35, opened 2015-10-20
State normalization for directives


vanessayuenn:
NgFor & NgIf first; other directives to be followed.

bertrandk:
These are the ones we're primarily concerned with:
- [ ] NgClass
- [ ] NgFor
- [ ] NgIf
- [ ] NgStyle
- [ ] NgSwitch
- [ ] NgSwitchWhen
- [ ] NgSwitchDefault

NgForm brings in another set:
- [ ] NgControlName,
- [ ] NgControlGroup,
- [ ] NgFormControl,
- [ ] NgModel,
- [ ] NgFormModel,
- [ ] NgForm,
- [ ] NgSelectOption,
- [ ] DefaultValueAccessor,
- [ ] NumberValueAccessor,
- [ ] CheckboxControlValueAccessor,
- [ ] SelectControlValueAccessor,
- [ ] NgControlStatus,
- [ ] RequiredValidator,
- [ ] MinLengthValidator,
- [ ] MaxLengthValidator

PR #36, opened 2015-10-20
state normalization - NgIf & NgFor

vanessayuenn:
Makes the directive states a bit more readable. Only NgIf and NgFor for now.
Should solve #35 .


bertrandk:
Looks good! :shipit:

PR #38, opened 2015-10-20
feat(adapter/controller): support deep tree

bertrandk:
The backend now supports deep nesting of components. It now also
supports traversing into components nested in the light DOM.

NOTE: This implementation assumes that a DebugElement is either a
component root or has child DebugElements from within the Light DOM.
Whether this assumption is true is worth investigating.


igor-ka:
looks good :shipit:

Issue #43, opened 2015-10-22
gulp build produces incorrect output on windows


Ristaaf:
When trying to build with gulp build, the transpiled typescript files gets put in app\app\ts this is because, gulp-rename does not normalize the directory names, so the line ` path.dirname = path.dirname.replace('app/ts', 'js');` will actually do nothing on windows, since path.dirname will have \ instead of / in the path.
By adding the line `path.dirname = path.dirname.replace('app\\ts', 'js');` after that one, everything works in windows too, this is a workaround though, since I think the real solution is to fix gulp-rename.

PR #44, opened 2015-10-22
Fixed build on windows 

Ristaaf:
by replacing app\ts with js in windows style dirs, and not only Linux and OSX style dirs according to issue #43 But a bit cleaner solution

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -5,6 +5,7 @@ var gulp = require('gulp');
 var rename = require('gulp-rename');
 var ts = require('gulp-typescript');
 var tslint = require('gulp-tslint');
+var path = require('path');
 
 gulp.task('default', ['build']);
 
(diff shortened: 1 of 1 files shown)


bertrandk:
Thanks for the PR @Ristaaf! 

Looking at the problem, it seems like we should use `path.join` on the substring path instead of directly specifying the path separator (i.e. `path.join('app', 'ts')`). Running two optimistic replacements sort of hides the intent, imo.

Ristaaf:
Agreed, there is some refactoring needed though, since the path used in gulp-rename is not the same as the node path that you get with require('path'). I will make a suggestion, and you can comment on that.

igor-ka:
Hi @Ristaaf thanks for the contribution. The code looks good to me, however I am assigning this to @bertrandk as he reviewed its originally.

bertrandk:
:+1: Looks great. Thanks a bunch @Ristaaf.

Issue #34, opened 2015-10-20
Show typescript sources in chrome devtools stacktrace


alexeagle:
If I include sourcemaps in an app, I can open the original typescript source file and set breakpoints.
However, stacktraces still reference the .js file so clicking through those links takes me to the wrong place.
Probably a chrome issue but it seems like you guys could drive the fix?

alexeagle:
Can post a minimal repro if that's helpful.

bertrandk:
From what I can tell, this is an issue with Chrome.

It may be worth looking into https://github.com/stacktracejs/stacktrace-gps or https://github.com/novocaine/sourcemapped-stacktrace when we take a stab at error notification in the extension.

/cc @winkerVSbecks

igor-ka:
Thank you for your feedback @alexeagle 

Sourcemaps in chrome extension seems to be as easy as everywhere else. There is a PR that should address this.

winkerVSbecks:
With the new Webpack build system we have source-maps generating. I just need to link it to the file or make it inline.

alexeagle:
We were able to generate sourcemaps with webpack - as I said, we can set
breakpoints in TypeScript files. The issue is just with presentation of the
stack trace.

winkerVSbecks:
@alexeagle I merged PR with the Webpack build like an hour ago. Just tested break-points and `console.logs`. There are 2 levels to this:

1. Devtools where Batarangle is running
2. Devtools to debug Batarangle itself

In (1) the logs are linked to the `.ts` files. In (2) the logs don't seem to be linked to `.ts` files but, the break-points are. 

Can you confirm that?

alexeagle:
great, that looks good from the screenshot. I don't have time to fetch the latest and verify right now, am working on other stuff. Thanks!

PR #45, opened 2015-10-23
Add TS sourcemaps support

igor-ka:
Adding source maps support seems to be as simple as generating them during the build process. This is identical to what we do when developing pure web applications using TS.

Tested with @winkerVSbecks on our injected script, content script and the dev tools page and stepped through the TS code.

This PR is related to issue #34

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -5,6 +5,7 @@ var gulp = require('gulp');
 var rename = require('gulp-rename');
 var ts = require('gulp-typescript');
 var tslint = require('gulp-tslint');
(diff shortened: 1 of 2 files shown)


bertrandk:
@igor-ka check the conflicts and I'll take a look again.

igor-ka:
@bertrandk i'll wait for @winkerVSbecks to finish his webpack setup as it might render this obsolete.

igor-ka:
@bertrandk this is not needed anymore, @winkerVSbecks  took care of this through webpack now.

PR #46, opened 2015-10-23
Re-factored search

winkerVSbecks:
- Refactored the search functions
- Highlight results in the order:
  + exact name match
  + fuzzy match
- Added types to ComponentDataStore


igor-ka:
Looks good :shipit:

PR #51, opened 2015-10-27
Webpack Build Setup

### LINKED ISSUE #43: gulp build produces incorrect output on windows

winkerVSbecks:
- Restructured to the project
- Switched to webpack for build
- Removed gulp
- Added an editorconfig
- Using only @reactivex/rxjs. Updated the app to use the new API

Delivers #49 and should fix #43 

@bertrandk @vanessayuenn @igor-ka Quite a major restructuring here. Please review and test it out locally. Thanks


.editorconfig
```diff
@@ -0,0 +1,13 @@
+root = true
```
igor-ka: not sure if this is supported on vscode yet, might be in the pipeline, probably would be useful to mimic those settings in vscode configuration for now

README.md
```diff
@@ -11,7 +11,7 @@ If you want to contribute or need help getting started, join us on Slack by fill
 To develop this extension the following environment is used
```
igor-ka: This needs to be updated to reflect new npm scripts?

frontend/components/node-item/node-item.ts
```diff
@@ -33,10 +33,9 @@ export class NodeItem {
 
     // Listen for changes to selected node
     this.componentDataStore.dataStream
-      .pluck('selectedNode')
-      .distinctUntilChanged((selectedNode: any) => {
-        return selectedNode ? selectedNode.id : '';
-      })
+      .map(({ selectedNode }: any) => selectedNode)
+      .filter((selectedNode: any) => selectedNode && selectedNode.id)
+      .distinctUntilKeyChanged('id')
```
igor-ka: I get an error on this method, i think it is called distinctUntilChanged instead...am i missing something?

tsconfig.json
```diff
@@ -1,43 +1,20 @@
...
-    "./app/ts/batarangle/backend/adapters/angular2.ts",
-    "./app/ts/batarangle/backend/adapters/base.ts",
-    "./app/ts/batarangle/backend/adapters/event_types.ts",
-    
-    "./app/ts/batarangle/backend/controllers/base.ts",
-    "./app/ts/batarangle/backend/controllers/dom.ts",
-    
-    "./app/ts/batarangle/backend/batarangle-backend.ts",
-    "./app/ts/batarangle/backend/entry.ts"
+    "typings/tsd.d.ts",
```
igor-ka: I think the files configuration might need changing, i.e. removed this way the location of this file will be taken as the root of the project and everything inside will be resolved. I'll look into this separately.

tslint.json
```diff
@@ -29,7 +29,7 @@
     "no-string-literal": true,
```
igor-ka: how do we lint with this setup?

webpack.config.js
```diff
@@ -0,0 +1,148 @@
...
+    filename: '[name].js',
+    sourceMapFilename: '[name].js.map',
+    chunkFilename: '[id].chunk.js'
+  },
+
+  resolve: {
+    root: __dirname,
+    extensions: ['','.ts','.js','.json'],
+    alias: {
+      'rx': '@reactivex/rxjs'
```
igor-ka: How is this alias used?

igor-ka:
Also, system.config.js file can go at this point as well.

igor-ka:
@winkerVSbecks  I tried it earlier today, looks good, just some minor questions and comments with regards to how it works. I will be using this setup to work on my feature and see if anything else comes up. But so far so good.

package.json
```diff
@@ -28,17 +37,24 @@
     "immutable": "^3.7.5",
     "json-formatter-js": "^0.2.0",
     "reflect-metadata": "0.1.1",
+    "rimraf": "2.4.3",
```
igor-ka: Does this need to go into devDependencies instead?

package.json
```diff
@@ -28,17 +37,24 @@
...
     "rx": "^4.0.0",
-    "systemjs": "^0.19.3",
+    "ts-loader": "0.5.6",
```
igor-ka: devDependencies for ts-loader and tsd as well?

tsconfig.json
```diff
@@ -1,43 +1,20 @@
...
+    "noImplicitAny": false,
+    "removeComments": true,
+    "noLib": false,
     "emitDecoratorMetadata": true,
-    "module": "umd",
-    "outDir": "./app/js/"
+    "experimentalDecorators": true,
+    "sourceMap": true,
+    "listFiles": false,
+    "outDir": "dist"
```
igor-ka: seems like the outDir option is not really used, should we remove it? Or rename it to build to reflect where the bundle files actually go?

.editorconfig
winkerVSbecks: yea, is there a config file for VS Code that we can include meanwhile?

README.md
winkerVSbecks: I updated from gulp build/flow to npm start/build. But, needs a couple more updates.

.editorconfig
igor-ka: There is, I can look into mimicking those settings in vscode configuration file, maybe under a separate issue. I know you prefer not to deal with code ;)

frontend/components/node-item/node-item.ts
winkerVSbecks: there is both `distinctUntilChanged ` and `distinctUntilKeyChanged`. Maybe I'm on a newer version of `@reactivex/rxjs`. I'll switch back to `distinctUntilChanged`

tsconfig.json
winkerVSbecks: :+1:

README.md
igor-ka: Ohh yeah you are right I think its `npm run build` ?
I think it would be useful if we could document our npm script here as well with short descriptions, at least the main ones

tslint.json
winkerVSbecks: I'll do that next. Webpack has some pre-loader options. If that doesn't work then I'll just setup linting through CLI or gulp as a part of the watch script.

frontend/components/node-item/node-item.ts
igor-ka: It could be that it is part of alpha6 for @reactivex and might be part of the newer typings, if thats the case I think there is no harm in upgrading to version 6 in our package.json

webpack.config.js
winkerVSbecks: This came with from the seed project. I believe the idea is that you can import from `rx` instead of `@reactivex/rxjs`. I just realized that we still have `rx` as a dependency so, it might not work. I'll remove it and test it out.

tslint.json
igor-ka: :+1:

package.json
winkerVSbecks: Indeed it does and I'll remove `system.config.js` too

webpack.config.js
igor-ka: :+1:

package.json
igor-ka: :+1:

package.json
winkerVSbecks: Possibly. Webpack might require these to be dependencies. I'll check.

tsconfig.json
winkerVSbecks: :+1:

.editorconfig
winkerVSbecks: Yea, that works for me. I noticed they have a feature request to support `editorconfig`. Hopefully soon.

winkerVSbecks:
- [ ] Update README
- [ ] Remove `system.config.js`
- [ ] `distinctUntilKeyChanged` ➡ `distinctUntilChanged`
- [ ] `@reactivex/rxjs` alias
- [ ] TS Lint
- [ ] `devDependencies`
- [ ] `outDir`

winkerVSbecks:
@igor-ka updated to incorporate your feedback

igor-ka:
Looks sweet :shipit:

PR #53, opened 2015-10-29
TS Lint

winkerVSbecks:
- Enabled TS linting through the `tslint` loader
- Updated files to fix linting errors
- Changed `@reactiveX/rxjs` ➡ `@reactivex/rxjs` to fix web-pack warnings 

I should mention, there is one lint error:

```
1:113-120 This seems to be a pre-built javascript file. Though this is possible, it's not recommended. Try to require the original source to get better results.
 @ ./~/json-formatter-js/dist/bundle.js 1:113-120
```

As we've discussed, we'll be replacing this library with something custom soon.


igor-ka:
@winkerVSbecks just looked through the code, looks great. :shipit:

PR #54, opened 2015-10-29
Fix injection/inspection and messaging workflow

### LINKED ISSUE #47: Rework script injection and messaging workflow
Currently the our "backend" script gets injected on the onload event of the inspected apps document.
This results in a situation where if the dev tools page was closed and re-open a refresh button has to be pressed in order to get the script injected again.

Need to investigate and provide a better workflow in terms of script injection. React Dev Tools source code will probably have an answer to this problem.

igor-ka:
- Add support for dev tools close/open scenarios
 - Remove 1000ms delay requirement from front end
 - Add injected script cleanup to improve the memory profile
 - Rename user/backend action names/constants to reflect what is actually happening

Cleanup

 - Add exclude to tsconfig.json to ensure intellisense works
 - Fix @reactivex/rxjs vs. @reactiveX/rxjs warnings
 - Fix some tslint errors

This issues resolves #47 and adds some more cleanup. This will allow for close/open scenarios to works with our extension, as well as address the freezing tab issue that was happening after a while.


typings/tsd.d.ts
```diff
@@ -1,7 +1,7 @@
-/// <reference path="../node_modules/immutable/dist/immutable.d.ts" />
-/// <reference path="../node_modules/reflect-metadata/reflect-metadata.d.ts" />
 /// <reference path="chrome/chrome.d.ts" />
 /// <reference path="es6-promise/es6-promise.d.ts" />
 /// <reference path="filesystem/filesystem.d.ts" />
 /// <reference path="filewriter/filewriter.d.ts" />
 /// <reference path="webrtc/MediaStream.d.ts" />
+/// <reference path="../node_modules/immutable/dist/immutable.d.ts" />
```
winkerVSbecks: Do we need these? Shouldn't these be picked up from `node_modules` automatically?

igor-ka: I tried to get rid of these, but unfortunately we do. In short these typings were written in a way that does not seem to play well with this approach. I think that once immutable is updated we can get rid of these.

winkerVSbecks: :+1:

winkerVSbecks:
Tested locally and everything works as expected. :shipit:

PR #55, opened 2015-10-30
Feat/state normalization

vanessayuenn:
addresses #35

### DIFF
--- a/backend/entry.ts
+++ b/backend/entry.ts
@@ -4,10 +4,10 @@ import {Angular2Adapter} from './adapters/angular2';
 let channel = {
   sendMessage: (message) => {
     console.log('Inspected script sending: ', message);
-    return window.postMessage({
+    return window.postMessage(JSON.parse(JSON.stringify({
       type: 'BATARANGLE_INSPECTED_APP',
       message
-    }, '*');
+    })), '*');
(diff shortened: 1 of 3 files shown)


igor-ka:
@vanessayuenn the code looks good, can you squash your commits?
You can list individual things as bullet point underneath a main commit message.

backend/entry.ts
```diff
@@ -4,10 +4,10 @@ import {Angular2Adapter} from './adapters/angular2';
 let channel = {
   sendMessage: (message) => {
     console.log('Inspected script sending: ', message);
-    return window.postMessage({
+    return window.postMessage(JSON.parse(JSON.stringify({
```
igor-ka: @vanessayuenn could you provide some rationale behind this? I just want to understand the reasons behind this.

vanessayuenn: It's (a bit of hack) to strip out the methods of the object being passed via postMessage, since they would cause a DataCloneError on certain directive components. If there's a more legit way to do it, I'd love to know.

igor-ka:
looks good :shipit: 
otherwise, @vanessayuenn could you open an issue to investigate a clean way of dealing with DataCloneErrors?

Issue #48, opened 2015-10-26
Provide contribution guidelines


igor-ka:
Need to think of proper guidelines documents so as to make external contributions easy, and smooth out the on-boarding experience.

bertrandk:
As per #63:
- [ ] Add an MIT license
- [ ] Add commit message guidelines (as much as we care)
- [ ] Review that tslint is adequate and can be relied upon fully
  - I really want to avoid contributors having to revisit PRs for styleguide issues when the linter told them they were good
- [ ] Add actual, clickable, link to github issues (as the issue tracker)
- [ ] Expound on guidelines for contributions
  - just more details, especially with regards to how feature requests work
  - a solution to the "I'm new and don't even know where to get started" problem
    - the crux of the problem is that there's a lot of "out-of-band" stuff that occurs on projects (design docs, mental models, previous decisions, etc...)
    - we don't need a comprehensive solution in the PR but a dedicated "newcomer" slack/IRC channel and a paragraph directing them there might suffice

vanessayuenn:
This is done.

PR #63, opened 2015-11-05
Add contribution guidelines

igor-ka:
- Add contribution guidelines
 - Minor fixes in README.md

This PR resolves issue #48

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,5 +1,44 @@
-# Contributing
+# Contributing to Batarangle
+
+### Pull Requests
+
+The core team will be monitoring for pull requests.
+
+*Before* submitting a pull request, please make sure the following is done…
+
+1. Fork the repo and create your branch from `master`.
+2. If you've added code that should be tested, add tests.
+3. If you've changed APIs, development workflow or aspects of the architecrure, update the corresponding documentation.
+4. TypeScript linter is part of our build task, make sure your code lints before submitting your PRs.
+
+## Bugs
+
+### Where to Find Known Issues
+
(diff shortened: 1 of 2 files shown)


igor-ka:
@vanessayuenn @winkerVSbecks @bertrandk 
Assigned this to Bertrand, but i think it would be worth having a look just in case i missed something.

cc @yuri

winkerVSbecks:
Looks good to me. Not sure we need to include the coding style section. That stuff should just be enforced using tslint, editorconfig, etc. Also, we should get the CI up and running soon.

bertrandk:
I few things I'd like to add:
- [ ] Add an MIT license
- [ ] Add commit message guidelines (as much as we care)
- [ ] Review that tslint is adequate and can be relied upon fully
  - I really want to avoid contributors having to revisit PRs for styleguide issues when the linter told them they were good
- [ ] Add actual, clickable, link to github issues (as the issue tracker)
- [ ] Expound on guidelines for contributions
  - just more details, especially with regards to how feature requests work
  - a solution to the "I'm new and don't even know where to get started" problem
    - the crux of the problem is that there's a lot of "out-of-band" stuff that occurs on projects (design docs, mental models, previous decisions, etc...)
    - we don't need a comprehensive solution in the PR but a dedicated "newcomer" slack/IRC channel and a paragraph directing them there might suffice

igor-ka:
@bertrandk the license is already in the repo, do you want it to be added to the guidelines?

TsLint, I personally hate some of the rules that are in there, can we just pick something together that works?

bertrandk:
A link should suffice for the LICENSE. It feels odd not mentioning it at all. I can also take a stab at one in the comments here and let you review it when you're free.

bertrandk:
:ship: at will...

PR #64, opened 2015-11-15
docs(README): correct build command

juristr:


igor-ka:
@juristr thanks, looks good!

PR #72, opened 2015-11-23
feat: add test infrastructure

### LINKED ISSUE #68: Setup test infrastructure
_**Objective:** Allow local tests and enable CI integration for each PR and merge._

## Concerns
- Integrate juliangruber/tape-run or some other TAP-producing test framework
  - It MUST report the right 0/1 status code when running within a browser (headless or otherwise)
- We need some "sample" passing tests
- We can probably get away with testing on the Chrome Browser present in CircleCI, for now...

vanessayuenn:
- add dependencies for testing
 - use separate webpack config for packaging tests
 - add npm test command
 - include some tests (still need more!)
 - port over backend tests from before

This will resolve #68


package.json
```diff
@@ -23,12 +23,14 @@
     "remove-tsd-loader-typings": "rimraf node_modules/typescript-simple-loader/typescript-simple-loader.d.ts",
     "tsd-update": "npm run remove-tsd-loader-typings && tsd reinstall --overwrite",
     "postinstall": "npm run tsd-update && tsd install && tsd link",
-    "start": "rimraf build && webpack --watch"
+    "start": "rimraf build && webpack --watch",
+    "test": "webpack --config webpack.test.config.js && browserify build/test.js | node_modules/tape-run/bin/run.js | node_modules/tap-spec/bin/cmd.js"
```
bertrandk: :fuelpump: use `node_modules/.bin/<foo>` here

bertrandk: :checkered_flag: scratch that, use this instead: `"test": "webpack --config webpack.test.config.js && cat build/test.js | tape-run | tap-spec"`

package.json
```diff
@@ -23,12 +23,14 @@
...
   },
   "dependencies": {
     "@reactivex/rxjs": "5.0.0-alpha.6",
     "angular2": "2.0.0-alpha.45",
     "basscss": "^7.0.4",
+    "browserify": "^12.0.1",
```
bertrandk: per your own comment, should we remove this?

package.json
```diff
@@ -39,6 +41,10 @@
   "devDependencies": {
     "object-assign": "4.0.1",
     "rimraf": "2.4.3",
+    "tape": "^4.2.2",
+    "tap-spec": "^4.1.0",
+    "tape-run": "^2.1.0",
+    "tsify": "^0.12.2",
```
bertrandk: :put_litter_in_its_place:

webpack.test.bootstrap.ts
```diff
@@ -0,0 +1,2 @@
+let testContext = (<{ context?: Function }>require).context('./', true, /\.test\.ts/);
```
bertrandk: :confounded: Comment plz?

bertrandk:
Test failures hang the runner for some reason on everything except phantomjs. :confused:

bertrandk:
This is the diff between browserify and no browserify:
```
(function e(t, n, r) {
    function s(o, u) {
        if (!n[o]) {
            if (!t[o]) {
                var a = typeof require == "function" && require;
                if (!u && a) return a(o, !0);
                if (i) return i(o, !0);
                var f = new Error("Cannot find module '" + o + "'");
                throw f.code = "MODULE_NOT_FOUND", f
            }
            var l = n[o] = {
                exports: {}
            };
            t[o][0].call(l.exports, function(e) {
                var n = t[o][1][e];
                return s(n ? n : e)
            }, l, l.exports, e, t, n, r)
        }
        return n[o].exports
    }
    var i = typeof require == "function" && require;
    for (var o = 0; o < r.length; o++) s(r[o]);
    return s
})({
        1: [function(require, module, exports) {
```
Support for commonjs-style require? Could we add a polyfill?

webpack.test.bootstrap.ts
igor-ka: @bertrandk this is my code :) I pointed @vanessayuenn to it as an example.
This is equivalent of saying find all the test files, and call `require` on each dynamically so as to make sure they are all a big happy family within the bundle generated by tests

bertrandk: Yeah, she told me. I wonder if we shouldn't be able to modify this and get rid of our dependency on browserify. See [this comment](https://github.com/rangle/batarangle/pull/72#issuecomment-159345392) and [this one](https://github.com/rangle/batarangle/pull/72#issuecomment-159350027). Essentially...I'm wondering if there's a module system polyfill (even es6-loader) that we need to include in the test webpack.

igor-ka: Not sure what is the error that you are getting on hanging tests, but as long as it is 'require is not define' type of thing we shouldn't need browserify.

bertrandk:
More info - In browsers, we're getting a JSON error during the period it hangs:
```
Uncaught SyntaxError: Unexpected token o
```

webpack.test.bootstrap.ts
```diff
@@ -0,0 +1,4 @@
+// This is equivalent of saying find all the test files, and call require on each dynamically so as to make sure they are all within the bundle generated by tests.
+
+let testContext = (<{ context?: Function }>require).context('./', true, /\.test\.ts/);
+testContext.keys().forEach(testContext);
```
bertrandk: :new::heavy_minus_sign:

webpack.test.config.js
```diff
@@ -0,0 +1,39 @@
...
+
+  resolve: {
+    extensions: ['', '.ts', '.js', '.jsx'],
+    modulesDirectories: ['src', 'node_modules']
+  },
+
+  node: {
+    fs: 'empty'
+  }
+};
```
bertrandk: :new::heavy_minus_sign:

bertrandk:
:sheep: :it:

PR #85, opened 2015-12-21
chore: migrate batarangle to angular2 beta

vanessayuenn:
- update to angular 2 beta, and some other dependencies
 - switch from using kebab case to camel case for directives in templates
 - change locations of imports
 - use NgZone.run() instead of lifecycle.tick()
 - update webpack.config to hide certain errors that will be fixed in future angular 2 beta version
 - move vendor list in webpack.config to a separate file for easier maintenance


igor-ka:
looks good to me, :shipit:

PR #88, opened 2015-12-30
Upgrade example app

SethDavenport:
This PR attempts to do several things:

1. move from JSPM/SystemJS to webpack, since that's what we use for batarangle itself.
2. update to Angular 2 beta
3. fix the example so it actually works
4. correct README steps for connecting batarangle to an actual app.


SethDavenport:
Specific issues fixed:
- replaced template references to `ng-if`, `ng-for` with `ngIf`, `ngFor`.
- zapped a bunch of linter errors (tabs, trailing whitespace, etc.).

SethDavenport:
There is still one open issue: angular2 beta 0 runs in dev mode by default; this doesn't play well with the way we're using ImmutableJS in TodoStore.

The `ngFor` in TodoList's template is bound to a property whose getter is generating a ___new array each time it is called___ using `Immutable.toArray()`:
 
```typescript
  get todos() {
    return this._todos.toArray(); // Effectively a defensive copy.
  }
```

When bound to an `ngFor`, this causes sadness:

> EXCEPTION: Expression 'todoStore.todos in TodoList@11:14' has changed after it was checked. Previous value: '[object Object]'. Current value: '[object Object]' in [todoStore.todos in TodoList@11:14]

Binding it to a vanilla array which gets mutated with `.push` works fine.  But it's somewhat at odds with our FP philosophy. @igor-ka @e-schultz let's talk about a better approach.

SethDavenport:
Fixed the final issue by making TodoStore more flux-like.  In effect, the component keeps a 'view' copy of the TodoList and subscripts to change events on the store, instead of attempting to bind directly to it's immutable property.

This also makes updating and removing tasks much cleaner by enforcing one-way data flow.

example-apps/todo-mvc-example/source/app.ts
```diff
@@ -1,14 +1,22 @@
-import {bootstrap, bind} from 'angular2/angular2';
+// Load Global Styles
+import 'todomvc-common/base.css';
+import 'todomvc-app-css/index.css';
+
+import {bootstrap} from 'angular2/platform/browser';
```
igor-ka: I think it's probably better to import this like 'import { bootstrap } from 'angular2/bootstrap';
I believe it all comes from the same place, but this is in the official examples. Am i missing something else about this import?

example-apps/todo-mvc-example/source/components/todo-list.ts
```diff
@@ -8,31 +11,50 @@ import {TodoItem} from './todo-item';
...
+        </ul>
+      </section>
+    </section>`
 })
-export class TodoList {
+export class TodoList implements OnInit, OnDestroy {
+  public todos: Todo[] = [];
+  private _subscription: Subscription<Todo[]>;
 
   constructor(@Inject(TodoStore) private todoStore: TodoStore) {
```
igor-ka: Should not need @Inject here, as in the case of TypeScript angular will inject based on 'TodoStore' type anyway

example-apps/todo-mvc-example/source/components/todo-list.ts
```diff
@@ -8,31 +11,50 @@ import {TodoItem} from './todo-item';
...
-    this.todoStore = todoStore;
   }
 
-  addTodo($event, newtodo) {
-    if ($event.which === 13 && newtodo.value.trim().length) {
-      this.todoStore.add(newtodo.value);
-      newtodo.value = '';
+  public addTodo($event) {
```
igor-ka: I like to omit public for brevity as it is the default, but do not have a strong preference

igor-ka:
Thanks for a great contribution, great to see your name here @SethDavenport :)

example-apps/todo-mvc-example/source/components/todo-list.ts
SethDavenport: OK, I'll change it.

example-apps/todo-mvc-example/source/components/todo-list.ts
SethDavenport: You're right. Will fix.

example-apps/todo-mvc-example/source/app.ts
SethDavenport: `angular2/browser/bootstrap` Is what's used in the Angular 2.0 quickstart: https://angular.io/docs/ts/latest/quickstart.html

That said, `angular2/boostrap` also works.  I'll defer to your wisdom.

SethDavenport:
@igor-ka : made the changes you requested and rebased.

igor-ka:
@SethDavenport LGTM :shipit: 

PS: I am surprised there is no need for zone.js import either in one of the files or webpack.config.js let me know if you have uncovered something in that regard.

SethDavenport:
@igor-ka : spent some time discussing the zone thing with @vanessayuenn .

From what I've read, it seems like `angular-polyfills` imports the necessary `zone` libs, along with `es6-promise`.  In fact if I include both I get a console warning about `window.zone` already being defined.

However we're still hunting down a batarangle issue where it only detects the `<todo-list>` component, not any of its children.  I suspect this is an angular2-beta/batarangle incompatibility but I'll leave this PR up until we get to the bottom of it.

SethDavenport:
For the record, the batarangle issue coincides with this entry in the console, and it happens when the Angular2 tab is opened for the first time:

```
Uncaught TypeError: nativeElement.setAttribute is not a functionAngular2Adapter._emitNativeElement @ angular2.ts:133Angular2Adapter._traverseTree @ angular2.ts:109(anonymous function) @ angular2.ts:122forEach @ es6-shim.js:1107Angular2Adapter._traverseTree @ angular2.ts:121(anonymous function) @ angular2.ts:57forEach @ es6-shim.js:1107Angular2Adapter.setup @ angular2.ts:56(anonymous function) @ entry.ts:19__webpack_require__ @ bootstrap 11a0ead175d757e7a340?343e:19(anonymous function) @ bootstrap 11a0ead175d757e7a340?343e:39__webpack_require__ @ bootstrap 11a0ead175d757e7a340?343e:19(anonymous function) @ bootstrap 11a0ead175d757e7a340?343e:39(anonymous function) @ bootstrap 11a0ead175d757e7a340?343e:39
batarangle-backend.ts:41 Content script received:  Object {type: "BATARANGLE_INSPECTED_APP", message: Object}
```

It appears to be failing to insert the `batarangle-id` attribute on anything other than the  `<todo-list>` node.

igor-ka:
@SethDavenport I went with including zone.js explicitly on my own app, and things were mostly working, but will try angular2-polyfill. Thanks for clarification.

SethDavenport:
@vanessayuenn @igor-ka 

I tried hooking up batarangle to the app from the Angular2 quickstart (https://angular.io/docs/ts/latest/quickstart.html) using ...

```typescript
import { AppViewListener } from 'angular2/src/core/linker/view_listener';
import { DebugElementViewListener, inspectNativeElement } from 'angular2/platform/common_dom';
import { bind } from 'angular2/core';

bootstrap(AppComponent, [
  bind(AppViewListener).toClass(DebugElementViewListener)
])
.then(applicationReference => {
  const w: any = window;
  w.ng.probe = inspectNativeElement;
});
``` 

... and I got the same error as above.  So it looks like we have a batarangle<->Angular 2 incompatibility happening with the way we're hooking Batarangle up to the DOM.  Since this doesn't appear to be caused by this PR, I'm going to merge and file a separate issue.

PR #91, opened 2016-01-06
add script to package the extension to a .crx format for unofficial distributions

vanessayuenn:
- run `npm run pack` to generate `batarangle-packaged.crx' which then can be installed in Chrome directly
 - update manifest.json to match the version # in package.json


bertrandk:
:smile: Great!

:shipit:

PR #94, opened 2016-01-15
package.json - fix rxjs dependency

wulfsolter:
Fix dependencies missing from package.json

rxjs doesn't install with dep version `"rxjs": "^5.0.0-beta.0",` but needs `"rxjs": "5.0.0-beta.0",` without the caret.

url-loader@0.5.7 requires file-loader@*

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "crypto": "0.0.3",
     "es6-promise": "^3.0.2",
     "es6-shim": "^0.33.6",
+    "file-loader": "^0.8.5",
     "immutable": "^3.7.5",
     "json-formatter-js": "^0.3.0",
     "reflect-metadata": "0.1.2",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -25,20 +25,21 @@
...
   },
   "dependencies": {
-    "rxjs": "^5.0.0-beta.0",
     "angular2": "^2.0.0-beta.0",
     "basscss": "^7.0.4",
     "browserify": "^12.0.1",
     "crypto": "0.0.3",
     "es6-promise": "^3.0.2",
     "es6-shim": "^0.33.6",
+    "file-loader": "^0.8.5",
```
vanessayuenn: Could you please move this under devDependencies since it's only needed for webpack?

vanessayuenn:
Looks good aside from that one small thing. :)
Please rebase with master then it's good to go.

wulfsolter:
Done :) Thanks @vanessayuenn

PR #95, opened 2016-01-18
chore: make batarangle work with apps built with beta-1

vanessayuenn:
Batarangle now looks for the first dom element with ngid as the root, but that means multi-root applications won't be supported for now.


sumitarora:
@vanessayuenn we can remove the function `_rootSelector` if it's not being used and associated occurrences.

vanessayuenn:
Made some changes as per @sumitarora's comment.

sumitarora:
looks good :shipit: :thumbsup:

PR #108, opened 2016-01-25
search should be case insensitive

sumitarora:
Fixed issue: https://github.com/rangle/batarangle/issues/105


vanessayuenn:
lgtm :shipit:

PR #112, opened 2016-01-27
Highlight element on mouse over

sumitarora:


vanessayuenn:
The highlighter still shows up for components that aren't actually visible in the DOM.

vanessayuenn:
LGTM :+1:

PR #116, opened 2016-01-29
Feat/add description to tree nodes

sumitarora:


vanessayuenn:
This looks good to me, but since both @sumitarora and I have codes in here, we might want another pair of eyes to give it one last look. @bertrandk?

bertrandk:
lgtm :sheep:

PR #125, opened 2016-02-03
Adding build configurations and Upload to S3 script

sumitarora:


crxmake.sh
```diff
@@ -4,6 +4,11 @@
 # (This will not be needed for official distribution)
 # Based on https://developer.chrome.com/extensions/crx#scripts
 
+node -v
+npm -v
+
+npm run build
```
vanessayuenn: can we add this to npm prepack command instead of here?

circle.yml
```diff
@@ -1,3 +1,16 @@
...
+
 test:
   pre:
     - git grep --color TODO | cat
+
+deployment:
+  release:
+    branch: release-build
+    commands:
+      - ./crxmake.sh
```
vanessayuenn: `npm run pack` instead?

sumitarora:
done

PR #127, opened 2016-02-08
streamlining kitchen sink UI

sumitarora:


sumitarora:
@igor-ka just style changes for kitchen sink UI for demo

igor-ka:
@sumitarora LGTM :shipit:

PR #129, opened 2016-02-09
upgrade to beta.3

sumitarora:


backend/adapters/base.ts
```diff
@@ -42,10 +42,6 @@ export interface TreeNode {
   state: Object;
   input: Object;
   output: Object;
```
igor-ka: Were we populating lastTickTime before? If so, maybe it's worth keeping it?

sumitarora: nope it always returned 0 
```js
  _getComponentPerf(compEl: DebugElement): number {
    return 0;
  }
```

igor-ka: :+1:

igor-ka:
@sumitarora I got this branch and went through the kitchen sink example with this code. Everything works either better or as good as before. 

In terms of code I'll leave a few comments. Its up to you if you want to address any of those in a separate PR. 

I will be using this for the webinar tomorrow.

backend/adapters/angular2.ts
```diff
@@ -29,26 +29,64 @@ declare var ng: { probe: Function };
...
+
   setup(): void {
     // only supports applications with single root for now
     const root = this._findRoot();
-    this._traverseTree(ng.probe(root),
-                       this._emitNativeElement,
-                       true,
-                       '0');
+
+    // this._traverseTree(ng.probe(root),
```
igor-ka: I think we should remove the commented out code.

backend/adapters/angular2.ts
```diff
@@ -78,24 +111,36 @@ export class Angular2Adapter extends BaseAdapter {
   }
 
   _findRoot(): Element {
-    return document.body.querySelector('[data-ngid]');
+    let elements: any = document.querySelectorAll('body *');
```
igor-ka: should probably use const instead of let for `elements` and `debugElement`

sumitarora:
@igor-ka Yes I would do it in different PR with some refactoring and upgrades too :smile:

backend/adapters/angular2.ts
```diff
@@ -241,37 +292,35 @@ export class Angular2Adapter extends BaseAdapter {
...
-    //     if (provider instanceof DirectiveProvider) {
-    //       props[provider.displayName] = provider.metadata.events;
-    //     }
-    //   }
-    // }
+  _getComponentInput(compEl: any): Object {
+    let props = [];
+    if (compEl.providerTokens.length > 0) {
+      try {
+        const directiveResolver: DirectiveResolver = new DirectiveResolver();
```
igor-ka: dont need `: DirectiveResolver`, as the type will be inferred by the compiler

example-apps/kitchen-sink-example/package.json
```diff
@@ -7,7 +7,7 @@
   },
   "dependencies": {},
   "devDependencies": {
-    "angular2": "2.0.0-beta.1",
+    "angular2": "2.0.0-beta.3",
```
igor-ka: zone.js dependency needs to be upgraded to 0.5.11 as there is a warning when building and running it that is related

example-apps/kitchen-sink-example/source/app.ts
```diff
@@ -9,19 +9,25 @@ import {bootstrap} from 'angular2/bootstrap';
 import KitchenSink from './containers/kitchen-sink';
 
 // For Batarangle Integration
-import {AppViewListener} from 'angular2/src/core/linker/view_listener';
-import {DebugElementViewListener, inspectNativeElement}
-  from 'angular2/platform/common_dom';
+// import {AppViewListener} from 'angular2/src/core/linker/view_listener';
```
igor-ka: This is great, no need to modify your application to run it with Batarangle.
:+1:

igor-ka: But we should remove the commented out code

package.json
```diff
@@ -33,7 +33,7 @@
     "pack": "./crxmake.sh"
   },
   "dependencies": {
-    "angular2": "^2.0.0-beta.2",
+    "angular2": "^2.0.0-beta.3",
```
igor-ka: same zone.js dependency that needs upgrading

igor-ka:
@sumitarora as per my previous comment, LGTM to :shipit: if you want to address the comments in a separate PR.

PR #133, opened 2016-02-11
Fix/reload issue

### LINKED ISSUE #114: Batarangle state doesn't update on page refresh
Steps to reproduce:

1. Run the todo-mvc-example
2. Add some tasks
3. Open batarangle: the components are shown as expected.
4. Refresh the browser window

The tasks go away in the app (because the example app isn't persistent).  However batarangle still shows all the components from the previous run. 

Workaround: close and reopen dev tools.

vanessayuenn:
should fix #114


igor-ka:
@vanessayuenn LGTM :shipit:

PR #134, opened 2016-02-11
clear info panel on reload

sumitarora:


igor-ka:
@sumitarora LGTM :shipit:

PR #135, opened 2016-02-11
adding debounce to update component tree after all updates are done

sumitarora:


frontend/batarangle.ts
```diff
@@ -48,8 +49,10 @@ class App {
 
     this.componentDataStore.dataStream
       .map(({ componentData }: any) => componentData)
+      .debounce((x) => {
```
igor-ka: @sumitarora can't we just use ```.debounce(500)```?

sumitarora: `.debounce(500)` just work with event's not data stream with updated version

igor-ka: :+1:

igor-ka:
@sumitarora LGTM :shipit:

PR #136, opened 2016-02-16
updating version to beta.6

sumitarora:
should fix https://github.com/rangle/batarangle/issues/144


igor-ka:
@sumitarora LGTM :shipit:

vanessayuenn:
Beta 6 has updated its dependency for zone to `zone.js@0.5.14` :)

PR #137, opened 2016-02-16
expose selected element on console and make use of it using $a

sumitarora:
Should Fix https://github.com/rangle/batarangle/issues/143


vanessayuenn:
looks good! :shipit:

PR #158, opened 2016-02-18
scroll to current element on selection

sumitarora:
should fix https://github.com/rangle/batarangle/issues/148


vanessayuenn:
LGTM :ship:

PR #160, opened 2016-02-19
chore: upgrade from tsd to typings

### LINKED ISSUE #140: Remove tsd and use typings instead

vanessayuenn:
- switch to typings
 - remove webpack settings that are no longer needed
 - update readme
 - should resolve #140 

In case of any issue with the upgrade, run `npm run clean` then `npm install`.


sumitarora:
:shipit:

PR #162, opened 2016-02-19
Tree view

sumitarora:
should fix https://github.com/rangle/batarangle/issues/153


frontend/channel/backend-messaging-service.ts
```diff
@@ -22,6 +22,9 @@ export class BackendMessagingService {
     });
 
     this.backgroundPageConnection.onMessage.addListener((message: any) => {
+      if (message.from && message.from === 'content-script') {
```
vanessayuenn: Does this only occur on refresh?

sumitarora: yes on reloading the app

vanessayuenn: This actually happens every time any content loads, regardless of whether it's a refresh, or a new page load, or even if the application is one being inspected or not. Probably not ideal.

vanessayuenn:
I would think there can be a better approach to message passing on reload, but we can tackle that later down the road. :shipit:

PR #166, opened 2016-02-22
adding code back to rerender component tree

sumitarora:
should fix: https://github.com/rangle/batarangle/issues/165


vanessayuenn:
:shipit:

PR #169, opened 2016-02-24
fix build script

vanessayuenn:
- change crxmake.sh to accomodate new project structure
 - fix stylistic build warning


sumitarora:
LGTM :+1:

PR #173, opened 2016-03-02
adding check for component type exists

sumitarora:
should fix: https://github.com/rangle/batarangle/issues/172


vanessayuenn:
:shipit:

PR #174, opened 2016-03-02
updating version to beta.7

sumitarora:


vanessayuenn:
:shipit: 
note: we should probably switch this from tsd to typings too at some point...

PR #182, opened 2016-03-08
Replacing findRoot Element code with new api

sumitarora:
should fix: https://github.com/rangle/batarangle/issues/179


igor-ka:
@sumitarora LGTM

PR #184, opened 2016-03-09
Replaced mutation observer with root element zone changes

sumitarora:
Should fix: https://github.com/rangle/batarangle/issues/181 and https://github.com/rangle/batarangle/issues/164

### DIFF
--- a/src/backend/adapters/angular2.ts
+++ b/src/backend/adapters/angular2.ts
@@ -35,7 +35,6 @@ import {DirectiveResolver} from '../directive-resolver';
 import * as Rx from 'rxjs';
 
 export class Angular2Adapter extends BaseAdapter {
-  _observer: MutationObserver;
   _tree: any = {};
   _onEventDone: any;
 
@@ -70,7 +69,6 @@ export class Angular2Adapter extends BaseAdapter {
       this._emitNativeElement);
 
     this._trackAngularChanges(ng.probe(root));
-    // this._trackChanges(root);
   }
 
   _trackAngularChanges(rootNgProbe: any) {
@@ -137,11 +135,6 @@ export class Angular2Adapter extends BaseAdapter {
     };
   }
 
-  cleanup(): void {
-    this._removeAllListeners();
-    this.unsubscribe();
-  }
-
   _findRoot(): Element {
     const ngRootEl = getAllAngularRootElements()[0];
 
@@ -175,32 +168,6 @@ export class Angular2Adapter extends BaseAdapter {
     this.addChild(compEl);
   };
 
-  _trackChanges(el: Element): void {
-    this._observer = new MutationObserver(this._handleChanges);
-
-    this._observer.observe(el, {
-      attributes: true,
-      childList: true,
-      characterData: true,
-      subtree: true
-    });
-  }
-
-  _handleChanges = (mutations: MutationRecord[]): void => {
-    this.reset();
-
-    // Our handling of the change events will, in turn, cause DOM mutations
-    // (e.g setting)
-    this._observer.disconnect();
-
-    const root = this._findRoot();
-    this._tree = {};
-    this._traverseElements(ng.probe(root),
-      true,
-      '0',
-      this._emitNativeElement);
-    this._trackChanges(root);
-  };
 
   _getComponentChildren(compEl: any): any[] {
     return <any[]>compEl.componentViewChildren;
@@ -212,10 +179,6 @@ export class Angular2Adapter extends BaseAdapter {
 
   _getNativeElement(compEl: any): Element {
     return compEl.nativeElement;
-  }
-
-  _removeAllListeners(): void {
-    this._observer.disconnect();
   }
 
   _selectorMatches(el: Element, selector: string): boolean {
--- a/src/backend/adapters/angular2.ts
+++ b/src/backend/adapters/angular2.ts
@@ -24,7 +24,7 @@
  * Supports up to 2.0.0-beta-1
  */
 
-declare var ng: { probe: Function };
+declare var ng: { probe: Function, coreTokens: any };
 declare var getAllAngularRootElements: Function;
 
 import { TreeNode, BaseAdapter } from './base';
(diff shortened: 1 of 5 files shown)


src/backend/adapters/base.ts
```diff
@@ -117,5 +117,5 @@ export abstract class BaseAdapter {
 
   abstract serializeComponent(el: any, event: string): TreeNode;
 
-  abstract cleanup(): void;
+  // abstract cleanup(): void;
```
igor-ka: let's stay away from keeping commented out code, we have history with source control and can always go back if needed. I would remove it.

sumitarora: :+1:

src/backend/backend.ts
```diff
@@ -30,13 +30,6 @@ window.addEventListener('message', function(event) {
 
     if (event.data.message.message.actionType ===
```
igor-ka: the message.message is not a part of this PR, but this is a leftover from the days when I used to contribute to this :), we should unnest this sometime in the future.

src/backend/backend.ts
```diff
@@ -50,12 +43,16 @@ window.addEventListener('message', function(event) {
       const highlightStr = '[batarangle-id=\"' +
         event.data.message.message.node.id + '\"]';
 
-      (<HTMLElement>document.querySelector(highlightStr)).scrollIntoView();
+      const element: any = <HTMLElement>document.querySelector(highlightStr);
```
igor-ka: Do we need a cast to HTML element here? does querySelector return any or something else?

src/backend/controllers/dom.ts
```diff
@@ -69,7 +83,11 @@ export class DomController extends BaseController {
     }
 
     if (ch) {
-      ch.sendMessage({ type: 'model_change', payload: this.model});
+      this.callToRenderTree.next({
+        channel: ch,
+        message: { type: 'model_change', payload: this.model }
+      });
+      // ch.sendMessage({ type: 'model_change', payload: this.model});
```
igor-ka: I would remove the commented out code here.

src/backend/controllers/dom.ts
```diff
@@ -41,6 +44,17 @@ export class DomController extends BaseController {
     this.model = [];
     this.adapter = adapter;
     this.channel = channel;
+
+    this.callToRenderTree = new Rx.Subject();
+    this.callToRenderTree
+      .debounce((x) => {
+        return Rx.Observable.timer(250);
```
igor-ka: I would remove the curlies and the round brackets on the argument to keep it as a one liner

src/backend/controllers/dom.ts
```diff
@@ -19,6 +19,8 @@ import { AdapterEvent } from '../adapters/base';
 import { Angular2Adapter } from '../adapters/angular2';
 import { BaseController } from './base';
 
+import * as Rx from 'rxjs';
```
igor-ka: I would import only what is needed from rxjs as opposed to the whole thing

src/backend/adapters/angular2.ts
```diff
@@ -32,22 +32,48 @@ import { DirectiveProvider } from 'angular2/src/core/linker/element';
...
       '0',
       this._emitNativeElement);
 
-    this._trackChanges(root);
+    this._trackAngularChanges(ng.probe(root));
+  }
+
+  _trackAngularChanges(rootNgProbe: any) {
+    const ngZone = rootNgProbe.inject(ng.coreTokens.NgZone);
+    ngZone.onEventDone.subscribe(() => { this._onEventDone.next();});
```
igor-ka: curlies are not needed here and I would remove them

src/backend/adapters/angular2.ts
```diff
@@ -32,22 +32,48 @@ import { DirectiveProvider } from 'angular2/src/core/linker/element';
...
-
   _tree: any = {};
+  _onEventDone: any;
+
+  constructor() {
+    super();
+    this._onEventDone = new Rx.Subject();
+
+    this._onEventDone
+      .debounce((x) => {
```
igor-ka: I would make this a one liner as it is very simple

src/backend/adapters/angular2.ts
```diff
@@ -32,22 +32,48 @@ import { DirectiveProvider } from 'angular2/src/core/linker/element';
 import { Description } from '../utils/description';
 
 import {DirectiveResolver} from '../directive-resolver';
+import * as Rx from 'rxjs';
```
igor-ka: I would import only what is needed from rxjs here

src/backend/adapters/angular2.ts
```diff
@@ -24,7 +24,7 @@
  * Supports up to 2.0.0-beta-1
  */
 
-declare var ng: { probe: Function };
+declare var ng: { probe: Function, coreTokens: any };
```
igor-ka: I am wondering if the typings for those are defined somewhere within angular's generated *.d.ts files? This way we could get rid of the ambient declaration here, but this is just a nice to have :)

igor-ka:
@sumitarora overall LGTM minus the comments. Otherwise I am not sure if I followed what is happening here 100% as I am a but out of the loop with regards to what API they provided us with regards to being notified of changes to the components.

I am giving it a :+1: , but let's have @vanessayuenn or @bertrandk give a second look through.
Otherwise this is great, good job!

src/backend/adapters/angular2.ts
bertrandk: Good call...I need to double check but for some reason I seem to recall that it was typed but the typing weren't exported. Probably has changed since this is a public API.

src/backend/backend.ts
bertrandk: I believe it returns Element instead of HTMLElement.

bertrandk:
The suspicion is that we're leaking a lot of DOM nodes in this routine. @sumitarora, did you plan on addressing that in this PR or treating it separately?

In any case, aside from the small cleanup advocated, this looks good to me.

src/backend/backend.ts
sumitarora: will do in this issue https://github.com/rangle/batarangle/issues/142

src/backend/controllers/dom.ts
sumitarora: using it in creating debounce observable

src/backend/adapters/angular2.ts
sumitarora: using it in creating debounce observable and subject

sumitarora:
@bertrandk I will do another PR for DOM issues

src/backend/adapters/angular2.ts
```diff
@@ -109,11 +133,6 @@ export class Angular2Adapter extends BaseAdapter {
     };
   }
 
-  cleanup(): void {
```
vanessayuenn: the new listener doesn't need to be disconnected/unsubscribed?

vanessayuenn:
only had one question about the unsubscribe, but it looks good otherwise. :+1:

src/backend/controllers/dom.ts
igor-ka: You can import `debounce` operators on its own as well.

src/backend/adapters/angular2.ts
igor-ka: You can import `debounce` operators on its own as well.

sumitarora: will do in another PR

src/backend/controllers/dom.ts
sumitarora: will do in another PR

src/backend/adapters/angular2.ts
igor-ka: :+1:

src/backend/controllers/dom.ts
igor-ka: :+1:

PR #186, opened 2016-03-11
chore: upgrade angular2 version to beta.9

vanessayuenn:
- upgrade both batarangle and kitchen sink app to angular2@beta.9
 - deprecate tsd and switch to typings


sumitarora:
:shipit:

PR #189, opened 2016-03-14
Info panel string updating

sumitarora:
Updating properties of type `string` or `boolean` was not possible from info panel this fixes that issue
To test open InputOutput part of kitchen sink and try updating any of isOn, isDisabled, turn properties of InputOutput component

### DIFF
--- a/src/frontend/components/info-panel/info-panel.ts
+++ b/src/frontend/components/info-panel/info-panel.ts
@@ -1,4 +1,4 @@
-import {Component, View, ElementRef, Inject} from 'angular2/core';
+import {Component, View, ElementRef, Inject, NgZone} from 'angular2/core';
 import {NgIf} from 'angular2/common';
 import * as Rx from 'rxjs';
 import {ComponentDataStore}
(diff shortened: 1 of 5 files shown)


src/frontend/components/info-panel/info-panel.ts
```diff
@@ -27,18 +27,20 @@ export class InfoPanel {
...
     private componentDataStore: ComponentDataStore,
     private userActions: UserActions,
+    private _ngZone: NgZone,
     @Inject(ElementRef) private elementRef: ElementRef
   ) {
 
     this.componentDataStore.dataStream
-      .filter((data: any) => data.action &&
-        data.action !== UserActionType.GET_DEPENDENCIES)
-      .debounce((x) => {
```
vanessayuenn: do we not need the `debounce` here anymore for performance sake?

sumitarora: No, we don't we fixed the batch call issue for which it was added in mutation observer one forgot to  remove this.

vanessayuenn: :+1:

vanessayuenn:
LGTM!

PR #197, opened 2016-03-14
Updating property from info panel if object

sumitarora:
should fix: https://github.com/rangle/batarangle/issues/191


vanessayuenn:
:shipit:

Issue #75, opened 2015-11-25
Investigate why browserify is needed for test to run properly


vanessayuenn:
Without browserify, failing tests cause indefinite hanging. (See comments in #72). Since the tests are already getting bundled by webpack, running it through browserify is redundant. Find out how we can take out browserify without tests hanging.

itanex:
@bertrandk What was the reason that this was deferred?

bertrandk:
Mostly a matter of priority @itanex given it's not very severe (zero user impact really...just an oddity). If you want to take a crack at it, I'd love the help. I'm also available if you have any questions.

itanex:
Gotcha. I assumed as much. I have a bunch of work on my plate now, this seems like fun to add to it. I will see what I can do.

mobinni:
I'm working on this issue and found a fix. Pull request will be out soon

mobinni:
This issue has been solved by PR #207

PR #207, opened 2016-03-17
Fix issue #75

mobinni:
[issue 75](https://github.com/rangle/batarangle/issues/75)

Regarding issue #75 - Piping tape-run requires the output of your build to be piped to stdout. By default webpack writes to file, the browserify command will pipe to stdout on file compilation. This is why you had to add it to be able to run your tests. My bugfix implements 'cat'-ing the build/test file so that it's sent to stdout. That way you can pipe your other commands. I removed browserify as a dependency. You can read about the differences between webpack and browserify that made me come to this conclusion here: https://gist.github.com/substack/68f8d502be42d5cd4942


sumitarora:
@mobinni LGTM works fine just update your PR with code in master :shipit: :rocket:

mobinni:
Updated with current master and ready for merge @sumitarora

PR #209, opened 2016-03-18
chore: upgrade to beta.10

vanessayuenn:
- upgrade angular2@2.0.0-beta.10
 - upgrade zone.js@0.6.4
 - prefer core-js polyfill over es6-shim and/or es6-promise
 - remove typings that are not needed
 - remove all @View annotations
 - workaround for temporary zone.js issue that surfaced with the beta10 upgrade (see https://github.com/angular/angular/issues/7660)
 - have webpack.test.config use the same vendor file for convenience

There are some known issues with beta 10, namely https://github.com/angular/angular/issues/7656 that cause the following build error:
```
ERROR in .../batarangle/node_modules/angular2/src/common/directives/ng_plural.d.ts
(2,10): error TS2305: Module '".../batarangle/node_modules/angular2/src/common/directives/ng_switch"' has no exported member 'SwitchView'.
```
Currently no workaround but it's not a blocking error.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,15 +24,16 @@
     "build": "npm run webpack --colors --display-error-details --display-cached",
     "webpack": "webpack",
     "clean": "rimraf node_modules typings",
-    "postinstall": "typings install",
+    "postinstall": "rimraf typings && typings install",
     "start": "rimraf build && webpack --watch",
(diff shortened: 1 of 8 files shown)


sumitarora:
:ship:

PR #222, opened 2016-03-23
Adding Router Tree Skelton

sumitarora:


vanessayuenn:
@sumitarora this can be shipped without the actual router tree implementation right?

sumitarora:
yes we just need to hide the div which displays router data

src/backend/adapters/angular2.ts
```diff
@@ -78,6 +79,18 @@ export class Angular2Adapter extends BaseAdapter {
...
 
+  showAppRoutes(): void {
+    const root = this._findRoot();
+    try {
+      const routes = ParseRouter.parseRoutes(
+        ng.probe(root).componentInstance.router.root.registry);
+
+      this.showRoutes(routes);
+    } catch (error) {
+      console.log(error);
```
vanessayuenn: what error is being handled here?

sumitarora: Any parsing errors or inconsistent data due to API changes. Don't want users to get blocked.

vanessayuenn: :+1:

vanessayuenn:
:100: :ship:

PR #226, opened 2016-03-24
Code cleanup

sumitarora:


src/frontend/actions/user-actions/user-actions.ts
```diff
@@ -73,6 +72,10 @@ export class UserActions {
     });
   }
 
+  /**
```
vanessayuenn: why removing all the comments?

vanessayuenn:
:ship:

PR #230, opened 2016-03-28
Fixed the multiple root node additions

sumitarora:
should fix: https://github.com/rangle/batarangle/issues/229


vanessayuenn:
:shipit:

PR #232, opened 2016-03-28
adding check on response

sumitarora:
Should Fix: https://github.com/rangle/batarangle/issues/231


vanessayuenn:
the original problem still persists even with this fix though
image

sumitarora:
This is seprate issue if its not angular2 app connection issue was coming in case of Angular2 apps also.

vanessayuenn:
Fair enough. I made #238 to track that issue. You can :ship: this one first. :+1:

Issue #238, opened 2016-03-29
Navigating away from Angular 2 apps with Batarangle opened triggers an error


vanessayuenn:
Steps to reproduce: 
1. open batarangle
2. navigate to another page that's not an angular 2 app
3. see error below
![image](https://cloud.githubusercontent.com/assets/6842965/14117776/344d77de-f5b4-11e5-8df2-f62285c42915.png)

That is because the backend script is getting injected on every page load (see [line 17-23 in `src/content-script.ts`](https://github.com/rangle/batarangle/blob/master/src/content-script.ts#L21). Probably should add some kind of check to see if there's an angular 2 app on the page.

mobinni:
Done

vanessayuenn:
Resolved by #250

PR #241, opened 2016-03-29
Bugfix/ip 238 angular2 check

mobinni:
Concerning #238 I switched out some imports to require statements for conditional loading. You check if your window object has the **ng** attribute. 

If so you're probably running Angular2 so you can start your adapter and hook into the DOM. 

I don't feel it's the cleanest fix in the world, since I prefer ES6 imports, but for the sake of low coupling I felt that backend.ts should be responsible for only loading **IF** ```window.ng``` is defined.

### DIFF
--- a/src/backend/backend.ts
+++ b/src/backend/backend.ts
@@ -1,5 +1,3 @@
-import {DomController} from './controllers/dom';
-import {Angular2Adapter} from './adapters/angular2';
 import Highlighter from './utils/highlighter';
 import ParseData from '../frontend/utils/parse-data';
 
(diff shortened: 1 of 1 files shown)


src/backend/backend.ts
```diff
@@ -14,10 +12,20 @@ let channel = {
...
+
+if (window.hasOwnProperty('ng')) {
+  const {Angular2Adapter} = require('./adapters/angular2');
+  const {DomController} = require('./controllers/dom');
+
+  adapter = new Angular2Adapter();
+  dom = new DomController(adapter, channel);
+}
+
+if(adapter && dom) {
```
sumitarora: Missing whitespace

src/backend/backend.ts
```diff
@@ -14,10 +12,20 @@ let channel = {
...
+  dom.hookIntoBackend();
+  adapter.setup();
+}
```
sumitarora: Trailing Whitespace

sumitarora:
@mobinni We can wrap second if inside the first one we don't need it separate other than that looks good. Also, Can you add a comment there why we are doing it like this?

mobinni:
@sumitarora done!

vanessayuenn:
Hey @mobinni, thanks a lot for the contribution! After reviewing your solution, I agree with your stance on preferring `import` over `require`. Here's what I have in mind in terms of a cleaner approach:
- Have `content-script` be responsible for determining whether `backend.js` should be injected.
- Create a separate script with a simple function to determine whether `ng` is available (similar to what you have already), if so, post a message for `content-script`.
- Inject said script to the page batarangle is trying to debug, just like how `backend.js` is injected.

This way _does_ require touching more parts of the application, so please DM me on slack if you have any question. :)

PR #247, opened 2016-03-30
chore: script injection

vanessayuenn:
get rid of `count = 0` and use a map instead, in case we need to inject other scripts.


sumitarora:
:shipit: :rocket:

PR #250, opened 2016-03-30
Feature/ng validate

mobinni:
Concerning #238 I switched out some imports to require statements for conditional loading. You check if your window object has the ng attribute.

Implemented approach suggested by @vanessayuenn in #241

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -13,12 +13,6 @@
     ],
     "persistent": false
   },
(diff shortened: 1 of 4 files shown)


sumitarora:
@mobinni looks good works as breeze :rocket:

vanessayuenn:
That looks great! :+1:

PR #254, opened 2016-03-31
fix: reload clears tree

vanessayuenn:
- currently the previous state of tree persists on navigating away to a non-ng2 application 
- solution: on reload/navigate to another page, sends message to clear the tree first.


sumitarora:
:ship: :rocket: :roller_coaster:

PR #258, opened 2016-04-04
Initialize if null and Fixing lint error

sumitarora:
Fixes: https://github.com/rangle/batarangle/issues/257


vanessayuenn:
:ship:

PR #259, opened 2016-04-04
Go to source file of the component

sumitarora:
Fixes: https://github.com/rangle/batarangle/issues/34

### DIFF
--- a/src/frontend/components/component-info/component-info.ts
+++ b/src/frontend/components/component-info/component-info.ts
@@ -41,7 +41,9 @@ export default class ComponentInfo {
     chrome.devtools.inspectedWindow.eval(
       evalStr,
       function(result, isException) {
-        console.log(result, isException);
+        if (isException) {
+          console.log(isException);
+        }
       }
     );
 
--- a/src/frontend/components/component-info/component-info.ts
+++ b/src/frontend/components/component-info/component-info.ts
@@ -32,6 +32,23 @@ export default class ComponentInfo {
     }
   }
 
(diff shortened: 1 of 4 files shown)


src/frontend/components/component-info/component-info.ts
```diff
@@ -32,6 +32,23 @@ export default class ComponentInfo {
...
+  viewComponentSource($event) {
+    const highlightStr = '[batarangle-id=\"' + this.node.id + '\"]';
+
+    let evalStr = `inspect(ng.probe(document.querySelector('${highlightStr}'))
+    .componentInstance.constructor)`;
+
+    chrome.devtools.inspectedWindow.eval(
+      evalStr,
+      function(result, isException) {
+        console.log(result, isException);
```
vanessayuenn: `result` and `isException` are mutually exclusive though

sumitarora: yes they are

vanessayuenn: so why are we logging them both at the same time?

sumitarora: This opens source panel for the file we don't want to do anything just display response error or result

vanessayuenn: i think we can just log the errors if there is any.

sumitarora: done

vanessayuenn:
LGTM :+1:

PR #274, opened 2016-04-06
color scheme! 🌈

vanessayuenn:
- put most of the colors in as variables
- monospace font! 😍
- change the default sans serif font to `Lucida Grande` to match the rest of chrome devtools


sumitarora:
:shipit: :100: :rocket:

PR #287, opened 2016-04-08
changing the change highlight color

sumitarora:


src/styles/variables.css
```diff
@@ -26,7 +26,7 @@
   --bt-link-color: #2828AB;
   --bt-editable-color: #2828AB;
   --bt-editable-highlight-color: #EBF2FC;
-  --bt-change-highlight-color: #FFF0F0;
+  --bt-change-highlight-color: #E7B4E4;
```
vanessayuenn: why? it doesn't go with the rest of the color palette from the style guide.

sumitarora: The previous one was way too light I talked to @justinachoi about it and this is the new color we came up with almost same just little darker.

vanessayuenn:
Well I guess if we have the designer's blessing! :ship: :it: :smile:

PR #288, opened 2016-04-08
Feat: dependency tab

vanessayuenn:
- remove old dependent components tab
- style accordingly

this needs to go out _after_ #281


sumitarora:
LGTM :ship:

PR #301, opened 2016-04-11
Adding changelog template and release process basic version

sumitarora:


igor-ka:
@sumitarora :+1: 
I'd like @bertrandk to have a look as well before we agree.

bertrandk:
Will take a look in a few hours if it's not pressing. Let me know if I need to prioritize @sumitarora.

sumitarora:
@bertrandk No Rush

vanessayuenn:
We will prob need two build script -- one for chrome store one for "nightly builds", cos chrome store only takes `.zip`.

sumitarora:
@vanessayuenn yes we would need that

bertrandk:
> - **master**: Always contains the code for the latest release. Code from `dev` branch must be merged into master before release.

I guess we should clarify that it's the latest **main** release. Seems like all bugfix (patch) releases occur on branches?

> * **Hotfix** - a bugfix release (typically from a branch) with version tag `vX.Y.Z` where Z is non-zero (e.g `v2.1.1`)

Does this mean we're releasing the HEAD of a branch or merging into master _from_ a branch that is not dev? I guess we'd then update dev? Are we better off cherry-picking commits from dev in that case?

> branch so `master` [has the]* code for release


Looks good to me overall. Is the plan to automate changelog generation and/or tag pushing. This is a very tedious thing to do manually and I think there are a few good npm packages that provide a CLI prompt generate changelogs from commits. We can also script the package upload.

In the Chrome store, we should probably first release to [test accounts](https://developer.chrome.com/webstore/publish#publishing-to-test-accounts) for sanity checks.

sumitarora:
Yeap we should release to test accounts first and I will be modifying the build process as well as change log generation.

PR #307, opened 2016-04-12
Renaming Batarangle to Augury in app

sumitarora:
Should fix: https://github.com/rangle/augury/issues/303

### DIFF
--- a/src/backend/adapters/angular2.ts
+++ b/src/backend/adapters/angular2.ts
@@ -121,7 +121,7 @@ export class Angular2Adapter extends BaseAdapter {
     const injectors = this._getComponentInjectors(debugEl, dependencies);
 
     description.unshift({
-      key: 'b-id',
+      key: 'a-id',
       value: id
     });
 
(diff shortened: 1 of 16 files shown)


src/backend/adapters/angular2.ts
```diff
@@ -121,7 +121,7 @@ export class Angular2Adapter extends BaseAdapter {
     const injectors = this._getComponentInjectors(debugEl, dependencies);
 
     description.unshift({
-      key: 'b-id',
+      key: 'a-id',
```
vanessayuenn: :+1: :smile_cat:

src/devtools/devtools.ts
```diff
@@ -3,7 +3,7 @@
  * on Chrome Development Tools window.
 */
 chrome.devtools.panels.create(
-  'Angular2',
+  'Augury',
```
vanessayuenn: do we want to keep this as `Angular 2`?

sumitarora: I don't think so, this was one of the feedback in usability study also. @igor-ka ?

sumitarora: @igor-ka Can you comment on this ?

bertrandk:
lgtm

PR #312, opened 2016-04-16
Bug: fix Angular 2 bootstrap import

juristr:
Fixes import of the `bootstrap` function which is exported by `angular2/platform/browser` now and not from `angular2/bootstrap`. Otherwise the project doesn't compile.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ Augury is a Google Chrome Dev Tools extension for debugging Angular 2 applicatio
 
 ## Supported Version
 
-Currently works with applications built in [Angular 2.0.0-beta.14](https://github.com/angular/angular/blob/master/CHANGELOG.md#200-beta14-2016-04-07) with _limited backwards compatibility_, which will change once Angular 2 stabilizes.
+Currently works with applications built in [Angular 2.0.0-beta.15](https://github.com/angular/angular/blob/master/CHANGELOG.md#200-beta15-2016-04-13) with _limited backwards compatibility_, which will change once Angular 2 stabilizes.
 
 ## Join Our Slack Team
 
--- a/package.json
+++ b/package.json
@@ -31,19 +31,19 @@
     "pack": "./crxmake.sh"
   },
   "dependencies": {
-    "angular2": "^2.0.0-beta.14",
+    "angular2": "^2.0.0-beta.15",
     "basscss": "^7.0.4",
     "core-js": "^2.2.2",
     "crypto": "0.0.3",
     "d3": "^3.5.16",
-    "es6-promise": "^3.0.2",
+    "es6-promise": "^3.1.2",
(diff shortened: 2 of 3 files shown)


sumitarora:
@juristr Can you update `package.json` and `Readme.md` to latest beta.15 also?

juristr:
@sumitarora Ok, will do later

juristr:
@sumitarora Btw, there seems to be also an issue with the Circle configuration, 'cause currently (without this PR being merged) the build fails..

image

..while the CircleIO build strangely passes. Didn't look into the details why that's the case, but it's not really representative.

sumitarora:
@juristr yeap Circle CI it is behaving bit strange I am looking into that. Thanks

juristr:
@sumitarora Ok, updated the angular deps to beta.15. I'm getting builds errors however from the latest master. Could you check whether that's the case also on your machine?

image

Looks like a webpack configuration issue, but I don't have the time right now to investigate further. It shouldn't have been introduced by this PR.

sumitarora:
@juristr Looking into that now

sumitarora:
@juristr I checked build seems to be fine. I also checked the CircleCI build is fine there also, just the tests are failing due to zone.js error.

```js
test.js  5.75 MB       0  [emitted]  test
   [0] multi test 40 bytes {0} [built]
 [837] . \.test\.ts 666 bytes {0} [built]
    + 913 hidden modules

WARNING in ./~/json-formatter-js/dist/bundle.js
Critical dependencies:
1:113-120 This seems to be a pre-built javascript file. Though this is possible, it's not recommended. Try to require the original source to get better results.
 @ ./~/json-formatter-js/dist/bundle.js 1:113-120
    Error: More tasks executed then were scheduled.
        at ZoneDelegate._updateTaskCount (http://localhost:57437/bundle.js:12602:25)
        at ZoneDelegate.invokeTask (http://localhost:57437/bundle.js:12573:28)
        at Zone.runTask (http://localhost:57437/bundle.js:12466:49)
        at ZoneTask.invoke (http://localhost:57437/bundle.js:12636:35)
```

Error is only occurring with `angular2: 2.0.0-beta.15` with `zone.js: 0.6.10` not with `angular2: 2.0.0-beta.15` with `zone.js: 0.6.8`. I am trying to find the root cause.

juristr:
@sumitarora Oh, I see. Well I took the angular 2 configuration from the [angular2-seed project](https://github.com/mgechev/angular2-seed/blob/master/package.json#L110) which is working for me, as it is always difficult to know the exact combination of angular 2 + zone + rxjs etc dep versions 😄 .

sumitarora:
@juristr Yes, same is with our starter projects its bit weird it's just happening for the test, not the actual project or sample app I use to check. Checking test configurations again.

PR #313, opened 2016-04-16
fix circleCI badge due to proj name change

juristr:
Just noticed that the CircleCI badge is broken, most probably due to the name change from batarangle->augury.


sumitarora:
@juristr There is already a PR for the name change in the project. It  contains the change for CircleCI badge name also. https://github.com/rangle/augury/pull/307/files

PR #328, opened 2016-04-21
fix: font

vanessayuenn:
- change Augury font to match the devtools font that was changed in chrome v50.x
- remove that 1px padding that was bothering me


sumitarora:
👍 💯 :shipit:

PR #330, opened 2016-04-22
implement circle CI work around according to support

vanessayuenn:
remove `node_modules` and `typings` before build


sumitarora:
🚀 🎢 💯

Issue #333, opened 2016-04-23
Augury doesn't display component directives information


sumitarora:
* Open Component Tree
* Select any Component 
* Directives Information is not shown for the component

PR #336, opened 2016-04-24
displaying component directives

sumitarora:
Should resolve: https://github.com/rangle/augury/issues/333


vanessayuenn:
LGTM :shipit:

PR #344, opened 2016-04-25
fix: reload issue

### LINKED ISSUE #305: Page refresh makes Batarangle to loose track
Angular2 tab in Chrome Developer Tools stops tracking the app after one or few page refreshes. Its just showing the last opened windows/properties and stops refreshing. I get empty left panel and whatever was shown last in the right panel but I can't expand/collapse any items there.
shot-2016-04-12-095214

If I close the whole browser tab and open it again, Batarangle works fine again, until a page refresh. If I close the Developer Tools before refreshing, Batarangle works fine again.

I'm using batarangle-484.crx with beta14

### LINKED ISSUE #340: TypeError: cannot read property 'injectors' of undefined
image
This would happen randomly and cause Augury to crash.

vanessayuenn:
- fire clear tree event
 - clear tree & selectedNode on reload
 - remove component store stream from `info-panel` as `selectedNode` should be passed down by `frontend`
 - hide injector hierarchy when no node is selected
 - fix #340 
 - fix #305


src/frontend/components/info-panel/info-panel.ts
```diff
@@ -31,25 +31,8 @@ export class InfoPanel {
 
   constructor(
     private componentDataStore: ComponentDataStore,
-    private userActions: UserActions,
-    private _ngZone: NgZone,
-    @Inject(ElementRef) private elementRef: ElementRef
-  ) {
-
-    this.componentDataStore.dataStream
```
vanessayuenn: @sumitarora Just confirming there's no reason why we need to care about the component store stream here right? Since we're already subscribed to it in `frontend.ts`.

sumitarora: @vanessayuenn yes I wanted to do the change but forgot about it thanks 😄

sumitarora:
LGTM 💯 :shipit:

PR #345, opened 2016-04-26
changing slack status

sumitarora:


bertrandk:
lgtm. 🐑 🐑 🐑 

Why the `wontfix` label?

vanessayuenn:
i was also wonder why the `wontfix` label was on everything

bertrandk:
It is a novel development strategy. 😉

sumitarora:
I thinks waffle adds it if there is no issue associated with PR

PR #346, opened 2016-04-26
Injectors issue

sumitarora:
Should Fix: https://github.com/rangle/augury/issues/340


vanessayuenn:
I had identified the issue to be the same as the reload issue and addressed it in PR #344. Is this still needed?

sumitarora:
yes it is related to that only, it's still better to add that check I missed that initally

vanessayuenn:
lgtm!

PR #357, opened 2016-04-28
chore: demo for ngconf

vanessayuenn:
- took it out of `angular directive`
- make a route have different route name & handler name
- add more di to demo


sumitarora:
💯

PR #369, opened 2016-05-02
Updated version numbers in kitchen sink

sumitarora:


igor-ka:
@sumitarora LGTM :shipit:

Issue #244, opened 2016-03-30
Add a githook to checks coding standards with tslint


sumitarora:
- Modify the Circle CI build to fail on PR, if all rules are not passed

PR #382, opened 2016-05-07
Added lint check on CI

sumitarora:
Should Fix: https://github.com/rangle/augury/issues/244


src/frontend/components/component-info/component-info.html
```diff
@@ -1,91 +1,58 @@
 <div class="flex flex-column overflow-scroll col-12">
```
igor-ka: @sumitarora this is not linting related is it?

src/frontend/components/component-info/component-info.ts
```diff
@@ -1,40 +1,39 @@
...
 export default class ComponentInfo {
   private node: any;
+  private selectDependency: EventEmitter<string> = new EventEmitter<string>();
   private propertyTree: string = '';
-  private _input: Array<any>;
 
   constructor(
-    @Inject(ElementRef) private elementRef: ElementRef,
-    private userActions: UserActions
+    @Inject(ElementRef) private elementRef: ElementRef
```
igor-ka: @sumitarora do we need the @Inject here? I would think that injection is happening based on type here anyway.

sumitarora: @igor-ka no we don't I think this PR have some extra code removing that.

igor-ka: 👍 I can't fin the extra code, but this is minor and we can fix it later.

sumitarora:
@igor-ka done fixed conflicts

igor-ka:
@sumitarora LGTM :shipit:

PR #383, opened 2016-05-08
Avoid auto compile and build TS on saving file changes for editors like VIM, Atom, etc.

andrejkn:
When using Atom or VIM, every time a `.ts` file is changed it triggers auto compile and build. This causes generating `.js` and `.js.map` of the modified `.ts` file, which is very counter productive.


sumitarora:
:shipit:

andrejkn:
@sumitarora I've updated this branch and it no longer has conflicts.

Issue #386, opened 2016-05-10
Augury not working


canda:
I just installed augury from the Chrome Web Store and I'm trying to use it, but everything is just blank (see screenshot). Can you help me start debugging this error?

![screenshot](https://i.imgur.com/zlZl143.png)

Sometimes, not allways, I'm getting getAllAngularRootElements is not defined. Anyways, It doesn't work whether the error shows up or not

```
angular2.ts:144 Uncaught ReferenceError: getAllAngularRootElements is not defined
Angular2Adapter._findRoot @ angular2.ts:144
Angular2Adapter.setup @ angular2.ts:55
(anonymous function) @ backend.ts:20
__webpack_require__ @ bootstrap 6968be3…?8b54:19
(anonymous function) @ bootstrap 6968be3…?8b54:39
__webpack_require__ @ bootstrap 6968be3…?8b54:19
(anonymous function) @ bootstrap 6968be3…?8b54:39
(anonymous function) @ bootstrap 6968be3…?8b54:39
```

damain:
I have the same error it worked at first after install then is showing blank even though my app is showing databound content in the webpage

damain:
If i restart the browser I can see the content in augury again but when the page refreshes it goes blank again

sumitarora:
@damain we have identified the issue and will be pushing the new build soon with the fix. Thanks!!!

sumitarora:
Fixed in RC1 support

akopchinskiy:
@sumitarora Not fixed. Does not work still. Latest version of the extension at the moment. Angular: 6.0.1
![image](https://user-images.githubusercontent.com/39277134/41407546-0bb1721a-6fd0-11e8-8b13-7fb89270b126.png)

devashish-rane:
Restart the whole browser..it will work

PR #388, opened 2016-05-10
set undefined if no component found on UI for same id

sumitarora:
Should fix: https://github.com/rangle/augury/issues/381and https://github.com/rangle/augury/issues/386


igor-ka:
@sumitarora LGTM :shipit:

PR #389, opened 2016-05-10
Updated parsing logic for release candidate

sumitarora:


src/backend/adapters/angular2.ts
```diff
@@ -192,42 +200,48 @@ export class Angular2Adapter extends BaseAdapter {
...
 
-    parameters.forEach((param) => dependencies.push(param.name));
+      parameters.forEach((param) => dependencies.push(param.name));
+    }
 
     return dependencies;
   }
 
   _getComponentInstance(compEl: any): Object {
     // fix could be undefined (are we grabbing the right element?)
```
igor-ka: @sumitarora is this comment still needed? or should we still return ```... || {}```?

src/backend/utils/description.ts
```diff
@@ -84,6 +129,17 @@ export abstract class Description {
     ];
   }
 
+  private static _getControlStatusDesc(instance: any): Array<Property> {
```
igor-ka: I think we should write a generic function that works for all the properties, instead of hardcoding the key names here. Let me know if I am missing something.

PR #392, opened 2016-05-12
html update

danielschifano:
This PR updates the HTML across the codebase. The majority of the changes consist of:

- Spacing consistency
- Structure consistency 
- Stacking attributes for readability

### DIFF
--- a/popup.html
+++ b/popup.html
@@ -1,11 +1,11 @@
 <html>
   <head>
     <style>
-      /*------- Modal Structure -------*/
+      /* Modal Structure */
       html, body {
-        padding: 0px;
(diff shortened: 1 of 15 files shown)


sumitarora:
@danielschifano Is there a way we can enforce these rules moving ahead. Also can you check `popup.html` also for the same.

danielschifano:
@sumitarora @vanessayuenn I think just following a guideline will be fine. We should have a lint-css file though.

sumitarora:
LGTM 💯

vanessayuenn:
LGTM2!

PR #396, opened 2016-05-16
Allow background color to span full width upon hover

danielschifano:
This PR allows for the background color to span the entire width of the container. Prior it stopped at the beginning of each line.

Preview Below:

screen shot 2016-05-16 at 12 19 47 pm


sumitarora:
LGTM :shipit:

vanessayuenn:
🚀

PR #405, opened 2016-05-20
feat: move header bar out of the component tree view

### LINKED ISSUE #404: Move the header bar
We're moving the header bar up to above the tabs with the intention of adding a settings button later on for switching between light and dark themes. The actual dark theme implementation is tracked in #378.

screen shot 2016-05-18 at 1 59 23 pm

vanessayuenn:
- move the header + search bar to above the tabs
 - fix the layout issue caused by moving header
 - fix a small bug in dependencies
 - two-way bind search query
 - clear & disable search when router tree view is selected

resolve #404


danielschifano:
Looks Good! :shipit:  @sumitarora what do you think?

sumitarora:
🚢

PR #410, opened 2016-06-05
Don't add null dependencies non existing names

Riron:
In case of optionnal dependencies, the `param` value can be null.
Thus, there is no `name` property and the following error is thrown :
`Uncaught TypeError: Cannot read property 'name' of undefined`


sumitarora:
@Riron Can you update from dev not the master.

Issue #415, opened 2016-06-07
Augury:----Does not render a component under component tree when i refresh the page. 


priyathorwat:
Everytime I need to do "ng-serve" and run the app then the component will be displayed in a component tree. But  Once I refresh the page, component tree is blank.The angular cli version that i am using is--- 1.0.0-beta.5.

sumitarora:
@priyathorwat Which version of Augury are you using? This happens with all the apps or the once using angular cli only?

priyathorwat:
yes it is the same behavior with all the apps. I have tested with sublime,brackets ,visual studio code with  angular-cli...how can i solve it?

sumitarora:
@priyathorwat Are you getting the same behaviour here https://augury.angular.io/demo/ Are you using the latest version `1.0.1` of Augury?

priyathorwat:
No  this link https://augury.angular.io/demo/ works fine. yes i am using the current version of augury

PR #419, opened 2016-06-10
wait for updating dom if bootstrap takes long time

sumitarora:
should fix: https://github.com/rangle/augury/issues/415


src/utils/ng-validate.ts
```diff
@@ -1,4 +1,35 @@
...
+  const MutationObserver =
+    /* tslint:disable */
+    window['MutationObserver'];
+    /* tslint:enable */
+
+  const eventListenerSupported = window.addEventListener;
 
+  return function (obj, callback) {
+    if (MutationObserver) {
+      let obs = new MutationObserver((mutations, observer) => {
```
igor-ka: @sumitarora `const` instead of `let`?

sumitarora: done

igor-ka:
LGTM :shipit: @sumitarora

Issue #378, opened 2016-05-04
No support for dark mode in Chrome


steveblue:
In Chrome, a developer can set the theme to dark mode. Augury should display in dark mode as well.

<img width="1440" alt="screen shot 2016-05-04 at 2 43 50 pm" src="https://cloud.githubusercontent.com/assets/1171943/15028565/b2fbc982-1206-11e6-86de-68ac6384b608.png">

sumitarora:
@steveblue what are you using for the dark theme?

steveblue:
The dark mode built into chrome dev tools settings. It was released in Chrome Canary a couple months ago and is now available in Chrome builds. 

<img width="418" alt="screen shot 2016-05-04 at 3 26 24 pm" src="https://cloud.githubusercontent.com/assets/1171943/15029827/97c51b54-120c-11e6-944f-359319115db3.png">

vanessayuenn:
@steveblue thats for the suggestion! We will consider having dark theme in the future.

vanessayuenn:
I don't think there currently is a programatic way to find out whether Chrome is in dark mode or not.

sumitarora:
@vanessayuenn yeap I tried that already it changes the class on the window but in Augury opens in an iframe so it doesn't have access to parent window

vanessayuenn:
Yup so this is deferred for now. If anyone wanna take a crack at it, feel free to do so.

steveblue:
This feature is awesome! @sumitarora @danielschifano @vanessayuenn @igor-ka TY very much for implementing this. 

I can finally feel like I'm in stealth mode!

![screen shot 2016-08-23 at 12 02 52 am](https://cloud.githubusercontent.com/assets/1171943/17917934/539b4dca-6975-11e6-851a-91c399b845f3.png)


<img width="1293" alt="screen shot 2016-08-23 at 9 02 46 pm" src="https://cloud.githubusercontent.com/assets/1171943/17917925/439817e6-6975-11e6-94c1-46f1c9abc04c.png">

PR #420, opened 2016-06-10
finishing the dark theme implementation and toggle

sumitarora:
should fix: https://github.com/rangle/augury/issues/378


igor-ka:
@sumitarora @vanessayuenn @danielschifano this looks good to me, but some of the code here is not my strong suite, i.e. the CSS part. I can give it a 👍 let's have a second pair of eye on this just in case.

@clbond can you have a look and give a 👍 if it looks good? I'll complete the review then.

clbond:
@igor-ka Absolutely, will do

src/frontend/components/header/header.html
```diff
@@ -1,42 +1,76 @@
...
+        <label for="light">Light Mode</label>
+      </li>
+      <li class="p3">
+        <input type="radio"
+          #darkTheme
+          [checked]="theme==='dark'"
+          (change)="themeChange('dark', darkTheme.checked)"
+          name="theme"
+          value="dark">
+        <label for="dark">Dark Mode</label>
```
clbond: Is it not possible to hook this up to the Chrome DevTools setting instead of duplicating it with our own setting?

src/frontend/frontend.ts
```diff
@@ -151,6 +158,11 @@ class App {
       this.searchDisabled = false;
     }
   }
+
+  themeChanged(newTheme) {
+    this.theme = newTheme;
+    // TODO: do stuff here to persist the theme selection
```
clbond: Does this mean I will have to re-select dark theme each time I reopen Chrome / Augury?

src/frontend/components/header/header.html
```diff
@@ -1,42 +1,76 @@
 <div class="flex flex-column flex-stretch col-12">
-  <header class="bg-darken-1 border-bottom py3 px3">
-    <h4 class="regular flex flex-center primary-color col col-4 m0">
+  <header class="flex flex-justify bg-panel border-bottom py3 px3 dark-theme">
```
clbond: I can't see where this .dark-theme style is defined - is it defined? It looks like the right thing to use is "dark" or "light", not "dark-theme" - or am I misunderstanding?

clbond:
Comments are above. Not sure if any of these issues need to be addressed before merging.

src/frontend/components/header/header.html
sumitarora: chrome's dev api doesn't expose the flag for that

src/frontend/frontend.ts
sumitarora: yes @clbond look into storing the values in local storage and reload it

src/frontend/components/header/header.html
clbond: I think I found a way to extract this setting. I will take a shot after this PR is merged

src/frontend/frontend.ts
igor-ka: @sumitarora let's do this in a separate PR. @clbond can you open an issue for the storing the value of this setting?

igor-ka:
@clbond let's create separate issues based in the discussion in the comments above. Otherwise, looks like @sumitarora addressed the code issue, so :shipit:

sumitarora:
done created new issue: https://github.com/rangle/augury/issues/423

Issue #394, opened 2016-05-14
'Loading...' after Canary update?


tycho01:
I just updated my Chrome Canary version to `52.0.2735.0` (64-bit), and it appears Augury is stuck on 'Loading...' for me now. I can't exclude other causes yet, but was wondering if anyone else had experienced something similar. If not, I'm clueless atm and this can be closed.

igor-ka:
Response on github issue is required

tycho01:
@igor-ka: Sorry, I'm a bit confused as to what you meant by response. Could you elaborate please?
I'm still experiencing this issue; neither browser restarts nor Augury reinstalls appear to be fixing it for me so far.

igor-ka:
@tycho01 sorry for the confusing response :)

We will investigate this issue, as we need to figure out if this is a Chrome Canary issue (which seems to work for us) or if this is something else.

It would be great if you could tell us which version of angular2 you are using as well. We will be releasing a version that supports RC1 very soon. Otherwise, I we only support the beta for now.

PS. Also your app needs to run in dev mode for Augury to work

st-clair-clarke:
I am experiencing the same problem with Canary (latest version) too. No such problem with Chrome.

sumitarora:
@tycho01 @st-clair-clarke In the latest update of the chrome `window.postMessage` has been deprecated this is the reason why it's not working see the message below. We will investigate if it's possible to support both the versions with the next Augury Release

> 'window.postMessage(message, transferables, targetOrigin)' is deprecated and will be removed in M54, around October 2016. Please use 'window.postMessage(message, targetOrigin, transferables)' instead. See https://www.chromestatus.com/features/5719033043222528 for more details.

tycho01:
Thank you for the update. Interestingly, I had been using Augury with ng2 RC1 in production mode. Only after the Canary update did I start experiencing issues. But this sounds hopeful; until resolved I'll try stable Chrome then. :)

clbond:
[...]

PR #425, opened 2016-06-13
#394: Fix Augury hanging on load on Chrome Canary

clbond:
Augury was not compatible with Chrome Canary, it would always give a “Loading…” indicator and hang there. Some analysis had been done and posted to Trello which speculated that this was a result of window.postMessage API changing, but in fact it was totally unrelated to that. It turned out to be a bug either in Angular 2 or in Chrome Canary which prevents recursive components from working correctly with templateUrl. (Or at least, prevents our particular recursive component - bt-node-item, from working.) I will create a plunkr demonstrating the issue and post the bug to the Angular project.

In the meantime, this change fixes Augury on Google Chrome.


sumitarora:
LGTM 👍  @clbond Can you add link to the issue you create on Angular here also.

taguan:
I encountered the same issue on my Angular 2 project (Chrome) and "fixed" it the same way than you.
@clbond did you open an issue at angular side ? I am interested by following it.
BTW, I think that issue is the same : https://github.com/angular/angular/issues/9627

Issue #435, opened 2016-06-16
graphs colors need to match the rest of the design in dark theme


vanessayuenn:
![image](https://cloud.githubusercontent.com/assets/6842965/16135650/1ad84c7a-33f3-11e6-89cb-6b31e2d3710d.png)

TODO: write more details here

PR #437, opened 2016-06-17
fixing the color update on graphs

sumitarora:
Should fix: https://github.com/rangle/augury/issues/435

### DIFF
--- a/src/frontend/components/router-tree/router-tree.ts
+++ b/src/frontend/components/router-tree/router-tree.ts
@@ -12,6 +12,7 @@ import * as d3 from 'd3';
 export class RouterTree {
 
   @Input() routerTree: Array<any>;
+  @Input() theme: string;
   treeConfig: any;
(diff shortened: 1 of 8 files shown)


src/frontend/components/router-tree/router-tree.ts
```diff
@@ -75,11 +76,13 @@ export class RouterTree {
       .attr('class', (d) => d.isAux ? 'node-aux-route' : 'node-route')
       .attr('r', 6);
 
+    const fillColor: string = this.theme === 'dark' ? '#A5A5A5' : '#000000';
```
clbond: Is there any way we could avoid hardcoding colors inside the JavaScript? I believe you can set svg fill from inside a css block

clbond:
:shipit: although it would be nice to avoid hardcoding colors in JavaScript. css supports setting svg fill color:

svg circle {
  fill: red;
}

Issue #436, opened 2016-06-16
settings menu should close when click outside of it


vanessayuenn:
![image](https://cloud.githubusercontent.com/assets/6842965/16135784/e0f4c7c6-33f3-11e6-8166-ff5a433bb119.png)

TODO: write more details here

clbond:
This is not fixed, reopening

ericjim:
I'll look into it.

ericjim:
#477 only address this issue partially. There are two edge cases that need to be addressed:
* Clicking on the `augury-header` does not hide the menu when the menu is revealed.
* Clicking on the scroll bars does not hide the menu when the menu is revealed.

PR #438, opened 2016-06-17
fixing the settings close issue

sumitarora:
Should fix: https://github.com/rangle/augury/issues/436

### DIFF
--- a/src/frontend/components/header/header.ts
+++ b/src/frontend/components/header/header.ts
@@ -40,12 +40,22 @@ export class Header {
     }
   }
 
+  resetTheme() {
+    this.settingOpened = false;
+    document.removeEventListener('click', this.resetTheme.bind(this), true);
+  }
+
+  openSettings() {
+    document.addEventListener('click', this.resetTheme.bind(this), true);
+    this.settingOpened = !this.settingOpened;
+  }
+
   themeChange(theme, selected) {
     if (selected) {
       this.theme = theme;
       this.newTheme.emit(this.theme);
     }
-    this.settingOpened = false;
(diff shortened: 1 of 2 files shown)


src/frontend/components/header/header.ts
```diff
@@ -40,12 +40,22 @@ export class Header {
     }
   }
 
+  resetTheme() {
+    this.settingOpened = false;
+    document.removeEventListener('click', this.resetTheme.bind(this), true);
+  }
+
+  openSettings() {
```
clbond: Since presumably these are not public APIs, is there any reason they couldn't be marked private or protected?

src/frontend/components/header/header.ts
```diff
@@ -40,12 +40,22 @@ export class Header {
...
+    document.addEventListener('click', this.resetTheme.bind(this), true);
```
igor-ka: @sumitarora why do we need to add an event listener for a click? I would think that angular's `(click)=...` in the template will take care of this? Am I missing something :)

sumitarora: @igor-ka I want the document click within the context of this component not as a whole.

src/frontend/components/header/header.ts
sumitarora: correct we can but I won't do it on one component we can create an issue to do it for all the components just to be consistent in the code

src/frontend/components/header/header.ts
igor-ka: Ahh got it

igor-ka:
@sumitarora LGTM :shipit:

PR #439, opened 2016-06-17
adding changelog

sumitarora:


vanessayuenn:
@sumitarora i think you're missing some close brackets :P otherwise lgtm

PR #444, opened 2016-06-17
Upgrade to Angular 2.0.0 RC2

clbond:


sumitarora:
Continued here: https://github.com/rangle/augury/pull/445

Issue #417, opened 2016-06-08
Router link description breaks Augury doesn't render component tree


sumitarora:
* Go to url http://run.plnkr.co/xpBjtrhnUMBMZX0y/#/base/app
* Open the dev tools
* Navigate to augury tab doesn't render the tree throws exceptions

![pasted image at 2016_06_08 02_55 pm](https://cloud.githubusercontent.com/assets/198247/15895275/6870e150-2d58-11e6-84bb-46ba1d1ab3f6.png)

PR #451, opened 2016-06-23
Fixing the error for the Router Link

sumitarora:
Should Fix: https://github.com/rangle/augury/issues/417


igor-ka:
@sumitarora LGTM :shipit:

PR #452, opened 2016-06-23
updating to rc3

sumitarora:
Should fix: https://github.com/rangle/augury/issues/450


igor-ka:
@sumitarora so there are no breaking changes between rc2 and rc3? Only the package version needed an update? 

LGTM :shipit:

sumitarora:
yes didn't saw any apart from forms being deprecated warnings

Issue #453, opened 2016-06-23
Ignore files created from build output


ericjim:
The output of the script `crxmake.sh` should be ignored by the `.gitignore` file.

PR #454, opened 2016-06-23
chore: ignore build script output

ericjim:
#### Changes
* Ignore build script output

#### Closes
* https://github.com/rangle/augury/issues/453


igor-ka:
LGTM :shipit: @ericjim

Issue #363, opened 2016-04-29
fix the `openedNodes` logic


vanessayuenn:
maybe rename `openedNodes` to `closedNodes` so it's not confusing 😅

PR #455, opened 2016-06-23
chore: rename openedNodes to closedNodes

ericjim:
### Changes
* Rename `openedNodes` to `closedNodes`

### Closes
* https://github.com/rangle/augury/issues/363

### Notes
* I used WebStorm to tell me all exact instances of `openedNodes`, and on a one-by-one basis used my best judgement to convert from one variable name to another.
* Use the build from below to test it manually, it seems to work with the [demo](https://augury.angular.io/demo/).

### Build
* ~~[augury.zip](https://github.com/rangle/augury/files/330791/augury.zip)~~ (1)
* [augury.zip](https://github.com/rangle/augury/files/330807/augury.zip) (2)

### DIFF
--- a/src/frontend/actions/user-actions/user-actions.ts
+++ b/src/frontend/actions/user-actions/user-actions.ts
@@ -87,13 +87,13 @@ export class UserActions {
    * Update the node state after re rendering the tree.
    * Select the previously selected node and
    * Preserve state of previously Closed Component.
-   * @param  {Object} options.openedNodes list of opened Nodes id's
+   * @param  {Object} options.closedNodes list of closed Nodes id's
    * @param  {Object} options.selectedNode currently selectedNode
    */
-  updateNodeState({openedNodes, selectedNode}) {
+  updateNodeState({closedNodes, selectedNode}) {
     this.dispatcher.messageBus.next({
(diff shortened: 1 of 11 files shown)


src/frontend/actions/user-actions/user-actions.ts
```diff
@@ -87,13 +87,13 @@ export class UserActions {
    * Update the node state after re rendering the tree.
    * Select the previously selected node and
    * Preserve state of previously Closed Component.
-   * @param  {Object} options.openedNodes list of opened Nodes id's
+   * @param  {Object} options.closedNodes list of opened Nodes id's
```
ericjim: @vanessayuenn I am guessing you also want me to change `opened Nodes` to `closed Nodes` in the comment doc string. I just want to be absolutely certain.

sumitarora: @ericjim yes, They should be changed in comments also

ericjim:
@sumitarora I think this is ready. I walked it through the demo app, and I couldn't find any issues caused by this, I am optimistic 😄

sumitarora:
:shipit: you don't need to attach the build I can just checkout your branch and build it 😄

PR #456, opened 2016-06-24
chore: bump Augury to version 1.0.3, and some 1.0.2 changelog tweaks

ericjim:


CHANGELOG.md
```diff
@@ -1,7 +1,12 @@
 # Change Log
 
-## 1.0.2
+## [1.0.3](https://github.com/rangle/augury/tree/1.0.3)
```
ericjim: This link currently doesn't exist @sumitarora

sumitarora: yes before release we will create tag then it will be there

sumitarora:
rebase with dev :shipit: 🚀

CHANGELOG.md
```diff
@@ -1,7 +1,12 @@
...
+* Update Augury to Angular 2.0.0-rc.3 ([450](https://github.com/rangle/augury/issues/450))
```
igor-ka: I think it should be updating Augury to support RC 3 applications or something like that instead

ericjim: Just for my understanding, Augury is built with RC3, and it also supports applications that are written in RC3. @sumitarora

sumitarora: yes it does

Issue #459, opened 2016-06-27
Color Settings popup should be consistent with theme


sumitarora:
![screen shot 2016-06-23 at 4 19 27 pm](https://cloud.githubusercontent.com/assets/198247/16383514/62775b12-3c53-11e6-8965-48dd0e471e5c.png)

ericjim:
Will something like this work? Note, I have scrollbars always present on my machine, so you'll see white stripes where overflow is present.
<img width="1437" alt="screen shot 2016-06-27 at 11 55 31 am" src="https://cloud.githubusercontent.com/assets/1329167/16386280/229ee7e8-3c5e-11e6-972b-4a7f95aeb778.png">

igor-ka:
@ericjim @sumitarora the white stripes do not work. I'd rather keep it the way it is for now if it means having white stripes in there

ericjim:
@igor-ka I will clarify. The whitestripes have always been there, they are only very visible as a result of a setting that I've enabled on macOS. This should not stop us from merging in this PR. @danielschifano can you comment on what things we can do to prevent the bars from appearing. My theory is that CSS is using `overflow` in some mode.

<img width="645" alt="screen shot 2016-06-27 at 4 38 27 pm" src="https://cloud.githubusercontent.com/assets/1329167/16394935/996e60e8-3c85-11e6-99f4-4f6eaa88b48a.png">

sumitarora:
@igor-ka As @ericjim mentioned they are the default scrolls in Mac on browser. I don't have them enabled which is default behavior I believe. If we need to fix the scrollbars we should create another issue specific to that.

igor-ka:
@sumitarora @ericjim I thought that the scrollbars are "side effects" of this code. Otherwise yes, another issue makes total sense. In which case, :shipit:

sumitarora:
Closing issue PR already merged: https://github.com/rangle/augury/pull/463

PR #463, opened 2016-06-27
chore: apply dark mode to color settings menu

ericjim:
#### Changes
* Color menu is now updated to reflect theme changes.

#### Closes
* https://github.com/rangle/augury/issues/459

### DIFF
--- a/src/frontend/components/header/header.html
+++ b/src/frontend/components/header/header.html
@@ -51,7 +51,7 @@
   </header>
 
   <form>
-    <ul class="setting-dropdown list-reset border border-dark rounded bg-silver absolute right-0" *ngIf="settingOpened">
+    <ul class="setting-dropdown list-reset border border-dark rounded bg-silver absolute right-0" [ngClass]="{'setting-dropdown--dark':theme==='dark'}" *ngIf="settingOpened">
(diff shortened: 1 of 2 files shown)


src/frontend/components/header/header.html
```diff
@@ -51,7 +51,7 @@ <h4 class="regular flex flex-center primary-color m0">
   </header>
 
   <form>
-    <ul class="setting-dropdown list-reset border border-dark rounded bg-silver absolute right-0" *ngIf="settingOpened">
+    <ul class="setting-dropdown list-reset border border-dark rounded bg-silver absolute right-0" [ngClass]="{'setting-dropdown--dark':theme==='dark'}" *ngIf="settingOpened">
```
ericjim: There are many ways of add/removing a css class in ng2, I like this one the best. With that said, I am open to other ways of achieving the same result.

igor-ka:
@ericjim LGTM :shipit:

src/frontend/components/header/header.html
sumitarora: @ericjim this is the way we have done it in application for black theme 👍

sumitarora:
:shipit:

PR #470, opened 2016-06-29
WIP - added example with new router

sumitarora:


example-apps/kitchen-sink-new-router/source/components/router/start.ts
```diff
@@ -0,0 +1,90 @@
...
+      component: StartMain
+    }, {
+      path: 'child',
+      component: StartChild
+    }, {
+        path: 'router-data2/:name/:message',
+        component: RouterData2
+      }]
+  }
+  // { aux: '/auxcomp', component: AuxComp },
```
igor-ka: Can we remove the commented out code?

sumitarora: Its still WIP will do once ready

ericjim:
I'll look into fixing the demo under `/start`

igor-ka:
I think that the example app update should happen in conjunction with the support for the new router. This way we can just update the kitchen-sink app w/o adding a separate folder.

sumitarora:
@igor-ka I added a different one so we can have both versions of router working and make Augury work with both of them. As I lot of apps are still using old router.

sumitarora:
Closing this continued with Augury update here: https://github.com/rangle/augury/pull/471

PR #471, opened 2016-06-30
feat: component router support

ericjim:
#### Changes
* Support component router on `Router Tree` graph, support for deprecated router still exists.
* Add control flow around deciding which router is present and which to parse for the graph.
* Still supports Aux routes
* Component router has no notion of specificity, it is removed from the description on rollover
* Remove `user-info-checkone.ts` since it uses a removed API in angular2.


src/backend/utils/parse-router.ts
```diff
@@ -84,3 +85,52 @@ export class ParseRouter {
     };
   }
 }
+
+export function IS_OLD_ROUTER_HACK(router) : boolean {
```
ericjim: Hacks need to look ugly so they are hard to pick on 😉

example-apps/kitchen-sink-new-router/source/components/router/start.ts
```diff
@@ -47,13 +51,18 @@ export default class Start {
 
   constructor(private router: Router) { }
 
-  gotoRouterData1() {
-    this.router.navigate(['/start/router-data1', 'Message from router']);
-  }
-
-  gotoRouterData2() {
-    this.router.navigate(['/start/router-data2',
+  goToRoute(index) {
```
igor-ka: This is just an example app, but the code here seems a but confusing to me? Especially the index part of it.

Why not call the router methods directly from the template markup?

sumitarora: I have changed it to single method just to be consitent

ericjim:
I've removed user-info-checkone.ts, just like in the deprecated router example app. Refer to: https://github.com/rangle/augury/pull/475/files

example-apps/kitchen-sink-new-router/source/components/angular-directives/hello.ts
```diff
@@ -0,0 +1,12 @@
+import {Component} from '@angular/core';
+
+@Component({
+  selector: 'hello',
+  inputs: ['msg'],
```
igor-ka: @sumitarora @ericjim I think that the preferred way is to use `@Input` decorator instead no?

sumitarora: @igor-ka It is but input and output metadata doesn't show up if use annotations there's already a issue for this https://github.com/rangle/augury/issues/472

example-apps/kitchen-sink-new-router/source/app.routes.ts
```diff
@@ -0,0 +1,13 @@
+import { provideRouter, RouterConfig }  from '@angular/router';
+
+import { KitchenSinkRoutes } from './containers/kitchen-sink';
+// import { StartRouter } from './components/router/start';
```
sumitarora: will do a cleanup on that again

PR #477, opened 2016-07-05
fix: hide theme menu if a click occured outside

ericjim:
#### Closes
* https://github.com/rangle/augury/issues/436


sumitarora:
@ericjim Its working in tree view and properties panel but not when click on header, popup itself and scrollbars.

ericjim:
This one remains elusive 😞, I'll revisit tomorrow, and see whether or not it makes it into the release. Not that big of a priority, I reckon.

PR #479, opened 2016-07-06
Add support for @Input() and @Output()

clbond:


src/backend/adapters/angular2.ts
```diff
@@ -316,24 +320,58 @@ export class Angular2Adapter extends BaseAdapter {
...
+
+      // Get inputs from @Input()
+      const propMetadata = Reflect.getOwnMetadata('propMetadata',
+        debugEl.componentInstance.constructor);
+
+      if (propMetadata) {
+        for (const key of Object.keys(propMetadata)) {
+          for (const meta of propMetadata[key]) {
+            if (meta.constructor.name === (InputMetadata as any).name) {
+              if (inputs.indexOf(key) < 0) { // avoid duplicates
```
igor-ka: @clbond what happens if the input is defined twice (as in, how does Angular 2 deal with it)? Once as part of the @Component decorator and then again as a class property decoration using @Input.

I am asking this question outside of the context of Augury.

clbond: Frankly I am not sure. I will look into that.

clbond: But for our code at least, if it is defined twice (once in both places), we won't return an array containing duplicates

igor-ka: That is good, but I would like to mimic the logic of Angular 2 here.

clbond:
Actually, this PR is incomplete. The input now shows up in the "Input" list but it has no value... I guess there is another spot where you are actually retrieving the value that uses the same logic?

src/backend/adapters/angular2.ts
sumitarora: Looks good @igor-ka @clbond Angular doesn't allows us to declare properties multiple times
`Output event 'result' defined multiple times`

sumitarora:
:shipit: 🚀

src/backend/adapters/angular2.ts
igor-ka: @sumitarora I am assuming you get that error at runtime? Also, when you get the error the application is not loaded, or is it more of a warning?

PR #481, opened 2016-07-07
chore: remove deprecated router ref from example

ericjim:
#### Changes
* Remove deprecated router reference.
* Some linting issues addressed


example-apps/kitchen-sink-new-router/source/components/form-controls/form2.ts
```diff
@@ -1,11 +1,9 @@
 import {Component} from '@angular/core';
 import {FORM_DIRECTIVES, Location} from '@angular/common';
-import {ROUTER_DIRECTIVES, RouterLink, RouteParams, Router}
```
ericjim: We forgot to remove this, the new example no longer has the old router as a dependency.

igor-ka:
@ericjim LGTM :shipit: pending checks passing

sumitarora:
:shipit:

PR #482, opened 2016-07-07
Adding 1.0.4 change log

sumitarora:


ericjim:
We should also update https://github.com/sumitarora/augury/blob/changelog-update/README.md

Perhaps something along the lines of:
```
Currently works with applications built in Angular 2.0.0-rc.4 using the Angular Component Router. Augury also has limited backwards compatibility with the Angular Deprecated Router, this will change once Angular 2 stabilizes.
```

sumitarora:
Will be done in another PR

ericjim:
LGTM

PR #486, opened 2016-07-12
enhancement: add ChangeDetectionStrategy OnPush. Clean imports.

ericjim:
#### Changes
* Add OnPush to `bt-node-item`, which is the element that gets recursively made in the component tree.
* Removed unused imports.

#### Notes
I notice a subtle performance gain, however, it could've just been the placebo effect. My research continues.

### DIFF
--- a/src/frontend/components/app-trees/app-trees.ts
+++ b/src/frontend/components/app-trees/app-trees.ts
@@ -1,6 +1,9 @@
-import {Component, Output, EventEmitter, Input}
-  from '@angular/core';
-import {NgClass} from '@angular/common';
+import {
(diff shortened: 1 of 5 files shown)


igor-ka:
@ericjim is there any improvement due to this?

ericjim:
@igor-ka I perceive some, but I am afraid it might be the placebo effect. Ideally we would have cold hard numbers, or have someone test my branch on their app. I can close the PR until I have some numbers.

igor-ka:
@ericjim yeah, let's discuss metrics etc. tomorrow and we can take it from there

PR #494, opened 2016-07-19
chore: build augury, create crx file, and move it to the artifacts folder in circleci

ericjim:
#### Changes
* Build the crx and zip file for Augury and place it in the Artifacts folder in CircleCI
* The uploaded files are viewed within CircleCI under the Artifacts tab. https://circleci.com/gh/rangle/augury/908#artifacts (this is the build for this PR)

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -7,6 +7,8 @@ machine:
 dependencies:
   pre:
     - rm -rf node_modules typings
+  post:
+    - npm run build; ./crxmake.sh
 
 test:
   pre:
--- a/crxmake.sh
+++ b/crxmake.sh
@@ -13,13 +13,16 @@ name="augury"
 files="manifest.json build src images index.html frontend.html popup.html"
 
 crx="$name.crx"
-if [ $CIRCLE_BUILD_NUM ]; then
-  crx="$name-$CIRCLE_BUILD_NUM.crx"
-fi
-
 pub="$name.pub"
 sig="$name.sig"
 zip="$name.zip"
+
+# assign build name to zip and crx file in circleci env
+if [ $CIRCLE_BUILD_NUM ] || [ $CIRCLE_ARTIFACTS ]; then
+  crx="$name-$CIRCLE_BUILD_NUM.crx"
+  zip="$name-$CIRCLE_BUILD_NUM.zip"
+fi
+
 trap 'rm -f "$pub" "$sig"' EXIT
 
 # copy all the files we need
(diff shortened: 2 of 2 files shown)


circle.yml
```diff
@@ -7,6 +7,8 @@ machine:
 dependencies:
   pre:
     - rm -rf node_modules typings
+  post:
+    - npm run build; ./crxmake.sh
```
ericjim: after `npm install` we build the zip and crx file.

crxmake.sh
```diff
@@ -13,13 +13,16 @@ name="augury"
...
-if [ $CIRCLE_BUILD_NUM ]; then
-  crx="$name-$CIRCLE_BUILD_NUM.crx"
-fi
-
 pub="$name.pub"
 sig="$name.sig"
 zip="$name.zip"
+
+# assign build name to zip and crx file in circleci env
+if [ $CIRCLE_BUILD_NUM ] || [ $CIRCLE_ARTIFACTS ]; then
```
ericjim: if we are in CircleCI we use the build number to label the crx and zip accordingly.

crxmake.sh
```diff
@@ -59,9 +62,16 @@ sig_len_hex=$(byte_swap $(printf '%08x\n' $(ls -l "$sig" | awk '{print $5}')))
 
 echo "Wrote $crx"
 
+# move crx to artifacts folder in circleci
+if [ $CIRCLE_ARTIFACTS ]; then
```
ericjim: Once the build is finished, we move the files to the artifacts folder, which is the folder that corresponds to the build of the current branch. The folder is accessible in CircleCi (for example: https://circleci.com/gh/rangle/augury/908#artifacts)

igor-ka: That's great @ericjim this brings us 95% there. Is there a way to automate having a direct link to crx file within the PR page intself (on Github). It's a long short but would it be possible to add a general comment to a PR with a link? Any other way would do. If not this is definitely good already.

ericjim: @igor-ka I spent a whole 10min investigating (and no dice), maybe I can spend a little bit more time?

igor-ka: @ericjim yeah, i think it's not worth it. We can keep it as a separate issue. Thanks.

igor-ka:
@ericjim LGTM :shipit:

ericjim:
I do have a possible solution, but it will require research. If I am successful, I'll add it as a separate PR.

PR #500, opened 2016-07-25
Fix new router example still using old router for inner child routes.

andrewlo:
Split router config into different files.


sumitarora:
:shipit: 🚀

PR #501, opened 2016-07-25
Use @Input and @Output decorators instead of component properties.

andrewlo:


example-apps/kitchen-sink-example/source/components/angular-directives/hello.ts
```diff
@@ -1,12 +1,13 @@
...
-  inputs: ['msg'],
   template: `
   <h4>
     Message: {{msg}}
   </h4>
   `
 })
-export default class Hello { }
+export default class Hello {
+  @Input() msg;
```
igor-ka: @andrewlo is it a string or something else? I would add the type.

sumitarora:
:shipit:

PR #502, opened 2016-07-26
Fixes #487 - exclude option elements from getting augury id

### LINKED ISSUE #487: Cannot read property 'id' of null
For some reason, nothing is showing up in the component section of Augury.
augury_err

It gave me these errors in the console: 
augury_err2

And this is where the error was thrown: 
augury_err3

My HTML is the same for 4 different components and the other 4 are evaluated correctly by Augury.

Am I doing something wrong?

Let me know if you need more info

**EDIT** I've discovered that this line of code in my HTML file is causing these errors in Augury
`<input [(ngModel)]="details.name" id="name" type="text">`
Any ideas?

Thanks,

andrewlo:
Option elements are excluded from being added to the compoment tree.

However, we must also exclude them from being numbered when assigning
augury-id's.

Otherwise, if we have option elements with other elements after it, we
will get undefined elements in the parent component's child array in the
component tree, since it assigns to the child array using the augury-id
as an index to the array.

Example:
If we have 4 `<option>` elements with a `<div>` after it, then the four
`<option>` elements have augury-id 0 to 3 and the `<div>` would get id 4.

When adding to the child array of the parent, the four `<option>`
elements are not added. The `<div>` after is added, but is added to index
4 in the child array. This creates undefined elements in positions 0-3 in
the array.

The front-end is then given the component tree and a JS error is thrown
when the user highlights the null node.

### DIFF
--- a/src/backend/adapters/angular2.ts
+++ b/src/backend/adapters/angular2.ts
@@ -102,8 +102,8 @@ export class Angular2Adapter extends BaseAdapter {
     if (compEl.children.length > 0) {
       compEl
       .children
+      .filter((child: any) => !this._isComponentExcluded(child))
       .forEach((child: any, childIdx: number) => {
-
         let index: string = idx;
         if (child.providerTokens.length > 0) {
           index = [idx, this._tree[idx]].join('.');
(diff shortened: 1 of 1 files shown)


src/backend/adapters/angular2.ts
```diff
@@ -190,9 +189,7 @@ export class Angular2Adapter extends BaseAdapter {
 
     if (isRoot) {
       return this.addRoot(compEl);
-    } else if (nodeName !== 'option') {
-      // skipping the option to improve performance 
-      // It adds no value displaying node elements
+    } else if (!this._isComponentExcluded(compEl)) {
```
sumitarora: Is this still required or just a precaution?

sumitarora:
:shipit:

PR #505, opened 2016-07-28
fixes #504 - Add trackBy functions to ngFor in component tree

### LINKED ISSUE #504: Add trackBy to component tree ngFor's to improve performance
The component tree is a tree of nodes where each node object has a unique ID field. Some nodes may have many children, and parts of the tree can be added or removed.

We can use trackBy on the ngFor iterating through the children to improve performance.

andrewlo:
Add trackBy functions to ngFor in component tree, using the unique
id field for each node.

This should improve performance when adding or removing nodes.

### DIFF
--- a/src/frontend/components/component-tree/component-tree.ts
+++ b/src/frontend/components/component-tree/component-tree.ts
@@ -45,4 +45,7 @@ export class ComponentTree {
     }
   }
 
+  trackById(index: number, node: any): string {
+    return node.id;
+  }
 }
--- a/src/frontend/components/component-tree/component-tree.html
+++ b/src/frontend/components/component-tree/component-tree.html
@@ -1,6 +1,6 @@
 <!-- BLOCK VIEW -->
 <main class="monospace col-12 p1">
-  <bt-node-item *ngFor="let node of tree"
+  <bt-node-item *ngFor="let node of tree; trackBy: trackById"
     [changedNodes]="changedNodes"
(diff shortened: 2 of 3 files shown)


ericjim:
LGTM, though I am interested in know what @sumitarora thinks.

igor-ka:
@sumitarora @ericjim @andrewlo I would like to understand why this works :)
I think I am probably missing something, but at the very quick and first glance it seems like it shouldn't. 

Let's get some clarity on that.

sumitarora:
@ericjim @igor-ka @andrewlo I tried on a small example with nested tree it seems to be initializing whole tree again even when using trackby. Here is the example https://github.com/sumitarora/angular2-examples

src/frontend/components/component-tree/component-tree.ts
```diff
@@ -45,4 +45,7 @@ export class ComponentTree {
     }
   }
 
+  trackById(index: number, node: any): string {
```
danielfigueiredo: This PR looks good to me
if I could say something about it would be that the trackById function is duplicated in two different classes, however not sure if there is a common place to put it
either way this ^ comment is just cosmetic, so I'd say :shipit:

sumitarora: @danielfigueiredo check my comment above ☝️

sumitarora:
:shipit:

PR #512, opened 2016-08-05
fixes #511

### LINKED ISSUE #511: When searching nodes in component tree, horizontal scroll bar covers selected node.
When searching nodes in the component tree, Augury tries to scroll the selected node into view. However, we don't take into account the horizontal scroll bar height, which can cover the selected node.

This occurs when the searched node is below the current viewport.

To reproduced, search for an item that is below the current view of the component tree. It scrolls down, but the horizontal scroll bar at the bottom covers the selected node.

andrewlo:
When scrolling down to reveal selected node, take
horizontal scrollbar height into account. Otherwise,
scrollbar covers the revealed node.


ericjim:
Works well 👍. LGTM

igor-ka:
:shipit:

sumitarora:
🚢

PR #518, opened 2016-08-08
fixes #517 - Run theme change in an Angular zone since the chrome

### LINKED ISSUE #517: When dark theme is selected, light theme is shown for a second when Augury starts
When dark theme is selected, light theme is shown for a second when Augury starts before dark theme is applied.

Ideally, the app starts in dark theme.

andrewlo:
storage sync function seems to happen outside the zone and
Angular needs to update the view after the theme loads.

This caused an issue where the dark theme would not render until
the component tree load caused a view update.


igor-ka:
@andrewlo LGTM :shipit:

Issue #520, opened 2016-08-09
Unescaped quote characters are causing sanitizeHtml to complain and do an extra loop while rendering component headings in the tree.


xorgy:

xorgy:
Addressed by #522

PR #522, opened 2016-08-09
Escape double quotes in node description so sanitizeHtml doesn't complain.

xorgy:
The corresponding error message is printed for every node description render when running in development mode; polluting performance profiles and decimating performance.

This change may also have a small positive impact on performance in production, as sanitizeHtml will do one fewer parse loop per render, though in practice this is fairly cheap.


xorgy:
This PR addresses #520

sumitarora:
🚢 :shipit:

PR #523, opened 2016-08-09
Validate that the message has a 'data' key before dereferencing message.

xorgy:
This was throwing an exception on load, and causing spurious updates.


igor-ka:
@xorgy can you create an issue for this with some rough steps on how to reproduce an exception being thrown?

xorgy:
It happens every time the content script starts. #532

igor-ka:
@xorgy LGTM :shipit:

Issue #530, opened 2016-08-12
Run Augury's Angular 2 instance in production mode.


xorgy:
Debug/Development mode has a performance impact which we should not incur for end users.

xorgy:
Resolved by #531

xorgy:
Resolved by #531.

PR #531, opened 2016-08-12
Disable Angular development mode when NODE_ENV !== 'development'

xorgy:
This addresses #530


sumitarora:
@xorgy Setting the prod env should be part of the prod build also? Like `"build": "cross-env NODE_ENV=production webpack",` ?

xorgy:
Setting NODE_ENV would need to be part of the build environment, yeah. I wanted to get input on where/whether to do that in this PR.

sumitarora:
@xorgy Any progress on setting node env. as part of build using `package.json`

xorgy:
@sumitarora Sorry for the delay, forgot to push this yesterday.

package.json
```diff
@@ -21,7 +21,8 @@
   },
   "homepage": "https://github.com/rangle/augury",
   "scripts": {
-    "build": "npm run webpack --colors --display-error-details --display-cached",
+    "build": "webpack --colors --display-error-details --display-cached",
```
igor-ka: @xorgy how come no `npm run` anymore prepending the `webpack`?

sumitarora:
@xorgy Looks good one last change to run in watch mode we use `npm start` please configure it to run in dev mode also. After that change 🚢 💯

package.json
xorgy: If just "webpack" is safe for `npm run webpack`, then "webpack --colors --display-error-details --display-cached" is safe for `npm run build`

xorgy:
@sumitarora Will-do.

package.json
igor-ka: Sounds good @xorgy thanks for the explanation

igor-ka:
@xorgy LGTM :shipit:

Issue #535, opened 2016-08-16
Display a conspicuous loading indicator.


xorgy:
Augury takes some time to start up, it would be nice to see an indication that it is loading which is more visible than the current black "Loading..." text.

clbond:
This has long since been addressed by https://github.com/rangle/augury/commit/dfa410b6545f24e3cef87af775b4e9651ffad13f

PR #538, opened 2016-08-16
Add a conspicuous standalone loading indicator.

xorgy:
I deliberately use only inline assets, since if I used any disk assets it would delay the loading further.

Addresses #535


sumitarora:
@xorgy Do we still need it else close the PR and issue? Thanks

igor-ka:
@sumitarora we agreed to close this one. @xorgy please close this PR.

Issue #543, opened 2016-08-22
Reduce stylistic and behavioural inconsistency with the Chrome Dev Tools proper.


xorgy:
Currently, some of the UI components do not completely match similar components in dev tools, which makes Augury feel out of place. It would be preferable if Dev Tools-style components behaved as they do in the rest of Dev Tools.

xorgy:
This issue is overbroad. Further issues should be more specific.

Issue #542, opened 2016-08-22
The UI hierarchy should have the component info panel on the same level as only the component tree.


xorgy:
Currently, the "Component Tree" and "Router Tree" tabs are on the same level in the UI as the component information panel. The tree mode tabs should appear to "contain" the combined tree and info panel for the component tree view, and only the router tree for the router tree view.

Issue #537, opened 2016-08-16
Allow users to resize the component info panel.


xorgy:
It should be possible to resize the component tree and editor panels.

PR #544, opened 2016-08-22
Implement resizable SplitPane component

xorgy:
- Move the component info panel under the top tab bar.
 - Fix tab component left margin.
 - Fix global background color to avoid a white flash during load.
 - Clean up frontend.ts imports.

Addresses #537 and #542; includes a small piece of #543.

I'm unsure of how/whether to write automated tests for this; the tests would effectively be a rote mirror of the code itself, and they would be considerably larger.

### DIFF
--- a/src/frontend/components/split-pane/split-pane.html
+++ b/src/frontend/components/split-pane/split-pane.html
@@ -18,8 +18,6 @@
      style="width: 6px; top: 0; bottom: 0; margin-right: -3px; cursor: ew-resize; z-index: 500; position: absolute;">
   </div>
   <div
-     #movecapture
-     (mousemove)="moveCaptureMouseMove($event)"
-     (mouseup)="moveCaptureMouseUp()"
+     #overlay
      style="z-index: 5000; position: absolute; cursor: ew-resize; top: 0; left: 0; right: 0; bottom: 0; display: none;"></div>
 </div>
--- a/src/frontend/components/split-pane/split-pane.html
+++ b/src/frontend/components/split-pane/split-pane.html
@@ -12,11 +12,13 @@
      class="col overflow-hidden flex mh100pct">
     <ng-content select="split-pane-secondary-content"></ng-content>
   </div>
+  <!-- These style properties are inline because they define the behaviour of the resizer bar, which is invisible. -->
   <div
      #resizer
      (mousedown)="resizerMouseDown($event)"
      style="width: 6px; top: 0; bottom: 0; margin-right: -3px; cursor: ew-resize; z-index: 500; position: absolute;">
   </div>
+  <!-- These style properties are inline because they define the behaviour of the overlay, which is invisible. -->
   <div
      #overlay
      style="z-index: 5000; position: absolute; cursor: ew-resize; top: 0; left: 0; right: 0; bottom: 0; display: none;"></div>
--- a/src/frontend/components/split-pane/split-pane.html
+++ b/src/frontend/components/split-pane/split-pane.html
@@ -1,0 +1,25 @@
+<div
+   #wrapper
+   (window:resize)="windowResized()"
+   class="flex flex-auto">
+  <div
+     #primary
+     class="col overflow-hidden flex mh100pct">
+    <ng-content select="split-pane-primary-content"></ng-content>
+  </div>
+  <div
+     #secondary
+     class="col overflow-hidden flex mh100pct">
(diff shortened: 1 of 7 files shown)


xorgy:
And of course, I've left the second patch here separated for some commentary. In order to get drag events while the cursor is outside of the window (very frustrating to use this otherwise), we need to bind the event listener to window. Ironically, IE and Firefox have an API for accomplishing this more cleanly, but it's not implemented in Chrome.

src/frontend/components/split-pane/split-pane.html
```diff
@@ -0,0 +1,23 @@
...
+  </div>
+  <div
+     #secondary
+     class="col overflow-hidden flex mh100pct">
+    <ng-content select="split-pane-secondary-content"></ng-content>
+  </div>
+  <div
+     #resizer
+     (mousedown)="resizerMouseDown($event)"
+     style="width: 6px; top: 0; bottom: 0; margin-right: -3px; cursor: ew-resize; z-index: 500; position: absolute;">
```
ericjim: I am curious, what was your reasoning for inline styles.

src/frontend/components/split-pane/split-pane.html
```diff
@@ -0,0 +1,23 @@
...
+  </div>
+  <div
+     #overlay
+     style="z-index: 5000; position: absolute; cursor: ew-resize; top: 0; left: 0; right: 0; bottom: 0; display: none;"></div>
```
ericjim: ^ditto

src/frontend/components/split-pane/split-pane.html
xorgy: None of these styles have any utility outside of their function in this module. I could spend time making little basscss nuggets for them; but it would only obscure their meaning, degrade performance, and increase the amount of code involved.

I opted for inline styles because they are functional styles.

src/frontend/components/split-pane/split-pane.html
xorgy: ditto ^_^

src/frontend/components/split-pane/split-pane.html
sumitarora: @xorgy I totally agree with you but I would still like to see the styles in CSS as seeing it in the code someone else might think we are using inline styles and start doing so I want to avoid that. It's not a big deal but just sticking to conventions is good.

src/frontend/components/split-pane/split-pane.html
sumitarora: Same here

src/frontend/components/split-pane/split-pane.ts
```diff
@@ -0,0 +1,141 @@
...
+  @ViewChild('primary') primaryElement : ElementRef;
+  @ViewChild('secondary') secondaryElement : ElementRef;
+
+  // TODO: store the initial secondary pane width in a preference.
+  secondaryWidth = DEFAULT_SECONDARY_WIDTH;
+
+  // State for resizing
+  mouseX : number;
+  bounds : ClientRect;
+  guardMouseMoved;
```
sumitarora: We can have them typed as `Function` I believe

xorgy: Unfortunately, TypeScript complains when this is typed `Function`, possibly due to a typing bug. It seems to have a rule for methods named `addEventListener` which requires the use of a higher level object with a `handleEventMethod()` on it. I think it is applying that rule incorrectly, and in any case the type it thinks addEventListener wants is _not_ the type that addEventListener actually wants; so I have to leave this untyped.

sumitarora: ahh thanks

sumitarora:
@xorgy looks good one last question shouldn't when resizing it should do relative resizing of both panes?

xorgy:
@sumitarora The metaphor is basically that the secondary pane retains its size when the window is resized. This is the same behaviour as the Chrome Dev Tools proper; in both horizontal and vertical split panes.

xorgy:
Excellent question :- )

sumitarora:
:shipit:

src/frontend/components/split-pane/split-pane.html
xorgy: Offline discussion concluded that I should include a comment above each inline styled element to clarify why the styles are inline.

PR #546, opened 2016-08-25
Massive refactoring of entire backend and most of the frontend

clbond:
This PR rewrites the entire backend and the majority of the frontend code in Augury. I apologize because it is a massive huge commit that no reasonable person would want to review all the way through. But since it was such a thorough refactoring, I was not able to submit discrete pieces of work in small PRs. Anyway here is what this PR changes:

 * Rewrites the communication mechanism between the browser and the extension
   1. Messages are no longer stateless and each side may respond to any messages they receive
   2. This 'response' functionality is used throughout the application now in order to support complex UI flows
 * Rewrites the entire backend
   1. Rewrite the backend in a way that makes it much more solid and resilient. I think we are less likely to see random bugs and errors as people use Augury on their applications, either large or small.
   2. Creates several new serialization mechanisms: binary msgpack transmission; custom 'meta object recreator' serialization that creates a function that is capable of reconstructing the original object, including any circular references
  3. This allows us to display state that previously was excluded from the UI (anything that could not be JSON.stringify'd)
 * Rewrites large chunks of the frontend
  1. Mainly just to clean up and refactor code that had become unwieldy
  2. But also to reduce bugs and make the application state a lot easier to understand

If you have an interest in this PR I strongly suggest cloning it and running the application, checking it out, checking for bugs, etc etc.


clbond:
Two things I did not finish implementing:
1. The Dependency tab in the component info window
2. Searching for routes in the Router Tree (but search in Component Tree works and it would be simple to finish the router search using the same control, in fact there's already a placeholder that throws a 'not implemented' exception)

But I don't think these are big enough issues to stop the PR. I would like to get it merged because it's such a drastic departure that the merge conflicts are going to be unbelievably difficult to deal with. Basically every line that gets changed between now and when this PR gets merged is going to conflict.

src/backend/utils/description.ts
```diff
@@ -1,90 +1,72 @@
+import {DebugElement} from '@angular/core';
+
 export interface Property {
   key: string;
-  value: string;
+  value;
 }
 
 export abstract class Description {
```
igor-ka: @clbond I don't think we are gaining anything from making this class abstract

xorgy:
😬

Issue #553, opened 2016-08-26
Components of the same type are collapsed/expanded together


igor-ka:
Component of the same type on the component tree get collapse and expanded together, if they are all on the same level within the tree. The problem manifests itself when the component is expanded or collapsed. 

Also, the state accordion shows the state for the first component of the type in the same level of the tree, no matter which component of the type is selected.

clbond:
@igor-ka I think this may fix it: https://github.com/rangle/augury/pull/560. Not absolutely sure since I haven't been able to reproduce it though (but I saw it happen on your machine)

clbond:
Can you try https://github.com/rangle/augury/pull/561/ and see if it fixes this issue? If it doesn't fix the problem, if you could find another app that reproduces the issue that would be awesome

clbond:
This has long since been fixed

PR #560, opened 2016-08-26
Stop propagation of expand events

clbond:
I think this should fix https://github.com/rangle/augury/issues/553 but I am not sure because I cannot reproduce it


igor-ka:
@clbond it does not seem to fix the above issue for me.

clbond:
Alright, sorry. I wish I could access your application and reproduce it. I'll close this PR

PR #569, opened 2016-08-29
Fix blank screen on first install / "show HTML elements" option being out of sync on first install

clbond:
Fix blank screen on first install / "show HTML elements" option being out of sync on first install


sumitarora:
👍

PR #573, opened 2016-08-29
Significant performance improvement due to removal of deep clone operations inside of JSON diff/patch

clbond:
In this PR I have created a copy of a library that we depended upon, https://github.com/Starcounter-Jack/JSON-Patch, made a fork of it, made significant performance improvements to it, and then replaced our dependency on that library with our local fork.

The biggest performance improvement comes from the removal of the deep cloning operations that were happening inside the JSON patch library. These were completely unnecessary in the context of our use-case and introduced a significant performance bottleneck.

### DIFF
--- a/src/utils/patch.ts
+++ b/src/utils/patch.ts
@@ -1,0 +1,284 @@
+// This code contains work that is copyrighted to Joachim Wester (2013)
+// and is licensed under the unrestrictive MIT software license.
+//
+// The original code is available at https://github.com/Starcounter-Jack/JSON-Patch
+//
+// This code has been forked and heavily modified for performance.
+
+function equals(a, b) {
+  switch (typeof a) {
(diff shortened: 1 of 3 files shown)


src/utils/patch.ts
```diff
@@ -0,0 +1,284 @@
+// This code contains work that is copyrighted to Joachim Wester (2013)
```
igor-ka: @clbond I am assuming that only the minimal code that is needed for Augury was forked into this file from the original repo?

clbond: Yup, the bare minimum

igor-ka: 👍

igor-ka:
@clbond LGTM :shipit: 

PS. consider submitting a PR to the original repo if it makes sense

PR #574, opened 2016-08-29
Remove npm debug log; add 'npm run dev' script

clbond:
* Remove a file that never should've been added
* Add an 'npm run dev' script that allows you to automatically rebuild when a file changes (when doing development)


igor-ka:
@clbond LGTM :shipit:

ericjim:
Wash away my sins. 🎉

PR #575, opened 2016-08-30
Clean up component-info panel, including hiding accordion elements that are empty, expanding Providers and Properties by default, and so forth.

clbond:
* Hide accordions that are empty
* Clean up component-info panel
* Add back some missing elements like Directives 
* Expand by default the Properties and Providers accordions since they are important and usually contain few elements


igor-ka:
@clbond LGTM :shipit:

PR #577, opened 2016-08-30
* Add back Dependencies accordion / dependency / 'change detection' view in info panel

clbond:
* Add back Dependencies accordion / dependency view in info panel
* Add back missing Change Detection section
* Change import statements to avoid importing Angular into
  content-script since it is completely unnecessary in there.
  The only places we use Angular is in backend and frontend,
  and backend runs in a separate context from content-script.
  (This change was also necessary for change detection display)

### DIFF
--- a/src/backend/backend.ts
+++ b/src/backend/backend.ts
@@ -6,9 +6,10 @@ import {
   MutableTree,
   Node,
   Path,
-  createTreeFromElements,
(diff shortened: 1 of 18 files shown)


xorgy:
I'm going to rerun the build, seems network conditions prevented it from working. The patch looks good to me. :- )

PR #580, opened 2016-08-30
Color and layout fixes and theming simplification.

xorgy:
- Always include the expander in the state tree layout so that keys at the same level always line up.
 - Always apply a theme class to <bt-app> to simplify theming.
 - Remove all app.css theme overrides.
 - Move all single-use PostCSS variable values to the reference site.
 - Remove all unused PostCSS variables.
 - Correct tab menu light theme styles.
 - Move the Router Tree outside of the <split-pane>, which belongs to the Component Tree.
 - Remove unused node-item styles.
 - Move some component-specific styles from base.css to appropriate component css files.
 - Fix light theme colours.
 - Use ViewChild and element refs in Header instead of querySelector and global ids.
 - Wrap theme selector menu radio labels around their corresponding radio inputs so that clicking them selects the corresponding radio button.
 - Remove unnecessary null check on event.target in Header.resetIfSettingOpened(), since every onclick event has a non-null target element.
 - Use .transparent instead of .invisible on non-expandable state editor nodes, to prevent mouse events from being emitted.


sumitarora:
@xorgy UI changes looks good few things `Show HTML elements` is not working

xorgy:
@sumitarora strangely, the CRX build seems to be of an older version. There may be some state which is preventing it from building the new version.

xorgy:
@sumitarora I just tested the CI CRX from this, and it is functioning as expected. The _Show HTML elements_ toggle is working as well.

igor-ka:
@xorgy LGTM :shipit:

PR #584, opened 2016-08-31
Remove "Children" panel from component-info since all of that information is available elsewhere ("Show HTML") and is a pointless performance bottleneck

clbond:


xorgy:
LGTM :shipit:

sumitarora:
:shipit:

xorgy:
Looks good, tried the CRX, works as expected.

PR #585, opened 2016-08-31
Charts cleanup

xorgy:
This PR introduces CSS styling for the Router Tree and Injector Tree, which eliminates the need to pass `theme` into InjectorTree, InfoPanel, RouterTree, and in turn AppTrees. This also eliminates a side effect whereby changing the theme would cause the Info Panel and Router Tree to re-render (throwing away local UI state).


clbond:
:shipit:

igor-ka:
🚀

PR #587, opened 2016-08-31
Clean up Triangle.svg and augury-logo.svg

xorgy:
This fixes some aliasing artifacts on the edge of the Augury iris logo when rendered in Chrome, which are especially apparent when it's rendered at a small size. The `Triangle.svg` asset was especially malformed; it was so ugly that I will (from now on) use it as an example of a poor-quality SVG; notably with signs of having been edited with two versions of Sketch and Adobe Illustrator.


clbond:
:shipit:

sumitarora:
🚢

Issue #578, opened 2016-08-30
Stack overflow exceptions thrown when the component tree is very deep.


clbond:
Use tail recursion to avoid stack overflow exceptions for component trees of large depth. To reproduce, open kitchen sink and then the Stress Tester. Then add a tree of depth 2000.

clbond:
Additional information on the various sources of stack overflows for very deep component trees:

 * Angular 2 itself. For very large component trees, Angular will fail inside the dependency injector because it too uses recursive algorithms for constructing its tree structures. But it currently handles about double the depth that our code does so we should at least bring our support for deep component trees to the same level as Angular 2.
 * Serializers (all 3: binary, recreator, JSON). When serializing trees or tree deltas we can hit stack overflows.
 * Constructing our tree representation in `transformer.ts`. This can be easily fixed to use a nonrecursive strategy.
 * JSON diff/patch. This third-party library uses recursive algorithms for comparing two structures and generating patches. We could get around this by comparing or patching multiple subtrees in several operations instead of the entire tree in one operation.

xorgy:
@clbond Time for some loops.

clbond:
Ya. We have 3 options:
* Convert all of the above (except third-party code) to iterative algorithms
* Refactor algorithms to take advantage of tail recursion optimization in V8
* Use a strategy like https://github.com/rangle/augury/pull/588/files to execute 'recursive' operations in a shallow call stack (but retain their essential 'recursive' nature)

xorgy:
Honestly, I think the straightforward approach (loops and stack arrays) is a good idea. Probably performs better as well, if we spend much computational time on it at all.

xorgy:
@clbond Do you consider this addressed? My understanding is that we have met or exceeded Angular's own component nesting limits, which sees like enough to me.

clbond:
No. It is not addressed. Even component trees that are not that deep blow up

xorgy:
Not a priority to solve this, looks like it doesn't blow up with any real application we've tried.

instantaphex:
I would love to use Augury.  Unfortunately I've never gotten anything besides maxiumum call stack size exceeded.  The application runs fine.

andrewthauer:
Closing due to inactivity. If this becomes an issue on a bigger scale we will consider re-opening and/or look at addressing.

PR #588, opened 2016-09-01
Refactor the object-recreation serializer to avoid deep recursion and stack overflows

clbond:
This resolves stack overflows associated with the serializer. It is connected to https://github.com/rangle/augury/issues/578 but is not a complete solution for that issue. Stack overflows still occur for very deep component trees. But this fixes one source of them.

### DIFF
--- a/src/utils/serialize.ts
+++ b/src/utils/serialize.ts
@@ -1,4 +1,6 @@
-/// The intent of this function is to create a function that is itself able to
+import {functionName} from './function-name';
+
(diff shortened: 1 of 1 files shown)


xorgy:
This looks good to me :- )

igor-ka:
@clbond the code LGTM, however, when i run the build the highlighting is strange.

xorgy:
Conversation between @clbond, @igor-ka, and myself reveals that the highlighting bug/behaviour @igor-ka was seeing on this branch was present without this PR.

PR #589, opened 2016-09-01
Use a transparent red tint instead of a solid bright red colour for search box errors.

### LINKED ISSUE #556: Search colours are not shown correctly in dark mode
When the dark mode is enabled the colour shown are wrong for the search input field

xorgy:
This resolves #556


sumitarora:
👍

PR #591, opened 2016-09-02
Relatively minor performance fixes

clbond:
* Revert the recursion change I made to the serializer as it performs way worse than the old implementation
 * Do not expand large arrays by default in the State view because it
   takes a long time to render them
(Note: There is a big performance fix coming around message transmission but this is not it)


sumitarora:
LGTM :shipit: Assuming CircleCI passes

clbond:
Closing. Going to resubmit as part of a different PR

PR #595, opened 2016-09-06
Port Augury itself to Angular 2.0.0 RC6.

### LINKED ISSUE #563: Support RC6

xorgy:
This resolves #563

### DIFF
--- a/src/frontend/components/accordion/accordion.ts
+++ b/src/frontend/components/accordion/accordion.ts
@@ -1,11 +1,10 @@
 import {Component, Input} from '@angular/core';
-import {NgClass} from '@angular/common';
 
 @Component({
   selector: 'accordion',
   template: require('./accordion.html'),
 })
-export default class Accordion {
+export class Accordion {
   @Input() private sectionTitle: string;
   @Input() private defaultExpanded: boolean;
 
--- a/src/frontend/components/header/header.ts
+++ b/src/frontend/components/header/header.ts
@@ -1,12 +1,10 @@
 import {
   Component,
-  NgZone,
-  EventEmitter,
-  Output,
   ElementRef,
+  EventEmitter,
   Input,
-  Query,
-  ViewChild
+  Output,
+  ViewChild,
 } from '@angular/core';
 
 import {Search} from '../search/search';
@@ -30,9 +28,6 @@ type Route = any; // TODO(cbond): use real Route type
   host: {
     '(document:click)': 'resetIfSettingOpened($event)'
   },
-  directives: [
-    Search,
-  ]
 })
 export class Header {
   @ViewChild('menuElement') private menuElement: ElementRef;
--- a/src/frontend/frontend.ts
+++ b/src/frontend/frontend.ts
@@ -1,5 +1,4 @@
 import {
-  CUSTOM_ELEMENTS_SCHEMA,
   ChangeDetectorRef,
   Component,
   NgModule,
@@ -354,8 +353,7 @@ const declarations = [
     UserActions,
     ViewState,
   ],
-  bootstrap: [App],
-  schemas: [CUSTOM_ELEMENTS_SCHEMA]
+  bootstrap: [App]
 })
 class FrontendModule {}
 
--- a/src/frontend/frontend.ts
+++ b/src/frontend/frontend.ts
@@ -1,4 +1,5 @@
 import {
+  CUSTOM_ELEMENTS_SCHEMA,
   ChangeDetectorRef,
   Component,
   NgModule,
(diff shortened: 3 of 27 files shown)


src/frontend/components/header/header.ts
```diff
@@ -29,31 +27,28 @@ type Route = any; // TODO(cbond): use real Route type
...
   @ViewChild('theme-menu-button') menuButtonElement: ElementRef;
 
-  @Input() private selectedTab: Tab;
-  @Input() private options: Options;
-  @Input() private tree: MutableTree;
-  @Input() private routerTree: Array<Route>;
+  @Input() selectedTab: Tab;
+  @Input() options: Options;
+  @Input() tree: MutableTree;
+  @Input() routerTree: Array<Route>;
```
clbond: Why remove the `private`s? This is private state that should not be exposed externally

xorgy:
This is work-in-progress.

Known issues:
 - Component names and details do not appear in the Component Tree. I'm not certain of the cause.

src/frontend/components/accordion/accordion.ts
```diff
@@ -1,11 +1,10 @@
 import {Component, Input} from '@angular/core';
-import {NgClass} from '@angular/common';
 
 @Component({
   selector: 'accordion',
   template: require('./accordion.html'),
 })
-export default class Accordion {
+export class Accordion {
```
igor-ka: 👍 :) not a fan of default export either

src/tree/transformer.ts
```diff
@@ -213,11 +213,11 @@ const getMetadata = (element: Source): ComponentMetadata => {
 };
 
 const getComponentDirectives = (metadata: ComponentMetadata): Array<string> => {
-  if (metadata == null || metadata.directives == null) {
+  if (metadata == null || metadata['directives'] == null) {
     return [];
```
clbond: Directives are no longer part of component definitions as of RC6. Is this for backward compatibility?

src/frontend/components/header/header.ts
xorgy: Left-over from debugging it, forgot to re-add the private qualifiers.

src/frontend/frontend.ts
```diff
@@ -43,36 +45,51 @@ import {
...
+import {RouterInfo} from './components/router-info/router-info.ts';
+import {RouterTree} from './components/router-tree/router-tree.ts';
+import {Search} from './components/search/search.ts';
+import {Spinner} from './components/spinner/spinner.ts';
+import {SplitPane} from './components/split-pane/split-pane.ts';
+import {StateValues} from './components/state-values/state-values.ts';
+import {TabMenu} from './components/tab-menu/tab-menu.ts';
+import {TreeView} from './components/tree-view/tree-view.ts';
+
 require('!style!css!postcss!../styles/app.css');
```
clbond: The convention elsewhere is to leave the .ts off the import statement

src/frontend/components/header/header.ts
clbond: Ahh. Thank you

src/frontend/components/header/header.ts
```diff
@@ -29,31 +27,28 @@ type Route = any; // TODO(cbond): use real Route type
...
 })
 export class Header {
   @ViewChild('theme-menu') menuElement: ElementRef;
   @ViewChild('theme-menu-button') menuButtonElement: ElementRef;
 
-  @Input() private selectedTab: Tab;
-  @Input() private options: Options;
-  @Input() private tree: MutableTree;
-  @Input() private routerTree: Array<Route>;
+  @Input() selectedTab: Tab;
```
igor-ka: @xorgy what's the reason for getting rid of private properties and methods in this class?

src/tree/transformer.ts
xorgy: To add to that, this doesn't work for RC5 applications unless they still use the deprecated declarations. In RC6 they are gone altogether as you say. We should find some other way to determine which directives are invoked in the template. I'll drop this block out and put a TODO in here.

xorgy: I was wondering whether we wanted to retain any RC4/deprecated RC5 support for this; but Igor clarifies that it's a non-goal.

I'm stubbing this out, we will need to implement this a different way in the future.

src/frontend/components/accordion/accordion.ts
xorgy: This also makes it consistent with all of our other `@Component`-decorated modules, so that the `@NgModule.declarations` imports look consistent.

src/frontend/frontend.ts
igor-ka: @xorgy yeah, let's get rid of .ts

xorgy: Fair enough, I hadn't noticed.

igor-ka:
@xorgy LGTM :shipit:

Issue #558, opened 2016-08-26
Component tree should be expanded to 3 levels deep by default


igor-ka:
By default the component tree should be expanded only 3 levels deep. Currently, all the nodes of the component tree are expanded fully.

clbond:
Resolved by commit https://github.com/rangle/augury/pull/597/commits/7241597cbbf4fa9b8ef9159d2eca287e71ab875a

xorgy:
Confirmed resolved by #597.

PR #597, opened 2016-09-06
Only expand tree nodes 3 levels deep by default

clbond:
Addresses issue https://github.com/rangle/augury/issues/558

### DIFF
--- a/src/frontend/actions/user-actions/user-actions.ts
+++ b/src/frontend/actions/user-actions/user-actions.ts
@@ -26,8 +26,13 @@ export class UserActions {
   ) {}
 
   /// Toggle the expansion state of a node
-  toggle(node: Node) {
-    switch (this.viewState.expandState(node)) {
+  toggle(node: Node, defaultState: ExpandState) {
+    let state = this.viewState.expandState(node);
+    if (state == null) {
+      state = defaultState;
+    }
+
+    switch (state) {
       case ExpandState.Collapsed:
         this.viewState.expandState(node, ExpandState.Expanded);
         break;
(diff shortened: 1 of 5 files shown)


igor-ka:
@clbond there is a small issue where in order to expand a node that is below level 3 in depth it is required to click twice on the arrow of that node.

The first click on the arrow only selects the node in the component tree. A second click on the arrow is required to expand it. Once a component has been expanded it can be both selected and expanded/collapsed with one click.

Its reproducible with a depth stress tester in the demo app.

clbond:
That wouldn't be caused by this change. If that is the case then it was that way before this pr

clbond:
Anyway I'll check it out in a moment

clbond:
@igor-ka Good catch. Addressed in commit c30dac4. Thank you

src/frontend/actions/user-actions/user-actions.ts
```diff
@@ -26,8 +26,13 @@ export class UserActions {
   ) {}
 
   /// Toggle the expansion state of a node
-  toggle(node: Node) {
-    switch (this.viewState.expandState(node)) {
+  toggle(node: Node, defaultState: ExpandState) {
+    let state = this.viewState.expandState(node);
+    if (state == null) {
+      state = defaultState;
```
igor-ka: @clbond can this be done with default parameters in the signature instead?

clbond: No, because the default expansion state changes depending on what level/depth the node is at. It must be passed from `NodeItem`.

xorgy:
:shipit:

sumitarora:
:shipit:

src/frontend/actions/user-actions/user-actions.ts
igor-ka: ohh right i missed that :) thanks!

PR #598, opened 2016-09-07
Default to Light theme / radio button should have default selection

clbond:


clbond:
This will be addressed in subsequent PR

PR #599, opened 2016-09-07
- Remove "Show HTML elements" and replace with 3-option menu in the s…

clbond:
- Remove "Show HTML elements" and replace with 3-option menu in the settings panel
- Default to previous component display behaviour ("Hybrid") where we only show components that have providerTokens (logic taken from old Augury source to ensure same behaviour)


clbond:
Hang on - don't review this yet, saving is broken

sumitarora:
:shipit:

clbond:
Thank you Sumit!

PR #602, opened 2016-09-07
Replace the recursive and iterative stress testers with an exponential...

xorgy:
… recursive stress tester.

This should help us track some performance characteristics which had eluded us.


igor-ka:
@xorgy LGTM :shipit:

PR #604, opened 2016-09-07
Resolve issues with changing state or emitting values

clbond:
This resolves an issue with updating properties and emitting new values


xorgy:
:shipit:

PR #608, opened 2016-09-07
Show the change detection strategy bare instead of in an accordion.

### LINKED ISSUE #607: Display change detection strategy in the component info header
Currently the change detection strategy is shown in a _Change Detection_ accordion. It is only one line's worth of information, so it should be displayed on its own without the accordion.

xorgy:
This resolves #607


clbond:
:ship: :it:

PR #612, opened 2016-09-08
Move @Input info to the state editor, remove the Inputs accordion.

### LINKED ISSUE #605: Integrate the Inputs accordion information with the state editor.
We would like to show `@Input()` decorations in the State editor to improve the semantic mapping between the source representation and the debug view.

xorgy:
This resolves #605


clbond:
Nice! Runs perfectly. Great job.

PR #615, opened 2016-09-08
Resolve issue with $a not being defined on regular elements (only components)

clbond:


igor-ka:
@clbond LGTM :shipit:

xorgy:
:shipit:

PR #616, opened 2016-09-08
Move @Output() to State view just like @Input() now is

clbond:


src/communication/message-factory.ts
```diff
@@ -154,7 +156,7 @@ export abstract class MessageFactory {
       messageResponseId: message.messageId,
       serialize: serialization,
       content: response instanceof Error ? null : prepare(),
-      error: response instanceof Error ? prepare() : null
+      error: response instanceof Error ? prepareError(response) : null
     });
```
clbond: This is not part of the @Output() change but it is important nonetheless. If an uncaught exception happens, it doesn't get serialized properly in some instances

xorgy: Thanks for noting, I was wondering why this line was changing again. ;- )

src/frontend/components/render-state/render-state.css
```diff
@@ -0,0 +1,56 @@
+.emitter {
+  position: relative;
+  display: flex;
+  width: 100%;
+  border-radius: 2px;
+  box-shadow: none;
+  -webkit-transition: all 0.4s linear;
```
xorgy: Prefixed rules are not necessary since we're only deploying in chrome, and as a stretch, WebKit browsers from this decade.

src/frontend/utils/input-output.ts
```diff
@@ -0,0 +1,3 @@
+export interface InputOutput {
+  [key: string]: {alias: string};
```
xorgy: Now that I look at it, since you're using `hasOwnProperty` you can just make it a map directly to string instead of refcell of string.

xorgy:
Looks mostly good aside from the nitpicks I've raised. If those are addressed then it's good to go in my book.

src/frontend/components/render-state/render-state.html
```diff
@@ -4,19 +4,37 @@
...
+    <span class="info-key" [ngClass]="{output: classification(k) === StateClassification.Output}">
       <div [ngClass]="{
         expander: true,
         rotate90: !expanded(k),
-        transparent: keys(state[k]).length == 0
+        transparent: keys(state[k]).length === 0,
+        'display-none': hideExpander(k)
       }"></div>
-      <span *ngIf="level === 0 && inputs[k]" class="primary-color">@Input(<span class="info-value" [hidden]="!inputs[k].alias">'{{ inputs[k].alias }}'</span>)&nbsp;</span>{{k}}:
+      <span [ngSwitch]="classification(k)" class="primary-color">
```
xorgy: You should be very careful with whitespace here, especially if there might be a mismatch to the equivalent whitespace in the StateValues template.

src/frontend/components/render-state/render-state.ts
```diff
@@ -40,14 +80,19 @@ export class RenderState {
...
-
-    $event.preventDefault();
-    $event.stopPropagation();
+  private classification(key: string): StateClassification {
+    if (this.level === 0) {
+      if (this.inputs.hasOwnProperty(key)) {
+        return StateClassification.Input;
+      }
+      else if (this.outputs.hasOwnProperty(key)) {
+        return StateClassification.Output;
```
xorgy: Maybe these could be queried separately. There is no case where a property name at level 0 can be both `@Input` and `@Output` so you might as well test them separately in the template.

src/communication/message-factory.ts
igor-ka: In the future if the specific change is unrelated, let's do a separate PR/issue for it. Let me know if i am missing something.

PR #618, opened 2016-09-08
Fix recursive serialization issue that Eric found

clbond:
This issue will not affect any users. But it will produce a nasty exception in the console if you try debugging Augury itself (which I am assuming very few people will)

### DIFF
--- a/src/frontend/frontend.ts
+++ b/src/frontend/frontend.ts
@@ -176,7 +176,7 @@ class App {
         break;
       case MessageType.TreeDiff:
         if (this.tree == null) {
-          this.connection.send(MessageFactory.initialize(this.options)); // request tree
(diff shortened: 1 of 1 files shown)


ericjim:
Very nice

ericjim:
So in other words, it's not the router tree causing the issue, it was the dev tools, debugging the router tree which raised the exception.

clbond:
No, it was caused by this flow:

* Open an Angular 2 application with Augury closed
* Do some stuff in the application that causes it to rerender itself a couple times
* Open Augury

But there is no way this exception could affect use of the application in any way, because the frontend will request a tree on ngOnInit regardless and the offending message -- getting a diff for a tree that doesn't exist, because Augury wasn't open -- will essentially be discarded.

sumitarora:
:shipit: 🚀

PR #620, opened 2016-09-08
Remember expand actions in "State" view, so that when someone selects…

clbond:
… another component and then comes back, their expand/collapse state is the same


sumitarora:
👍 :shipit:

src/frontend/components/render-state/render-state.test.ts
```diff
@@ -1,17 +0,0 @@
-import * as test from 'tape';
```
igor-ka: @clbond @sumitarora did we just get rid of unit test here? :)

PR #630, opened 2016-09-09
Remove lodash and also fix the 'constant updating' issue first seen i…n the Raymond James app (which was also related to cloning)

clbond:


igor-ka:
@clbond LGTM :shipit:

PR #634, opened 2016-09-11
Introduce state property metadata system + ability to manually emit values to rxjs Subject/AsyncSubject/BehaviorSubject/etc (as with @Output() event emitters)

clbond:
New property metadata system allows us to treat all rxjs subjects as "emittable" from the Augury UI.

We can now do this kind of thing in the State panel:
screen shot 2016-09-11 at 7 09 27 pm

Whereas before we had to render the same State object like so:
screen shot 2016-09-11 at 7 09 56 pm

This also fixes a bug with `@Input()` not appearing where it should

### DIFF
--- a/src/backend/backend.ts
+++ b/src/backend/backend.ts
@@ -3,9 +3,12 @@ import {DebugElement} from '@angular/core';
 import {Subject} from 'rxjs';
 
 import {
+  Metadata,
   MutableTree,
   Node,
   Path,
+  instanceWithMetadata,
+  serializePath,
 } from '../tree';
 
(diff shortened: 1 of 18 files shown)


sumitarora:
@clbond Does it also allow to emit the events and application respond to it?

clbond:
I don't understand the question

clbond:
It emits values the same way your application would `emit()` / `next()` them. The app cannot tell whether you manually emitted a new value from the UI or whether the application did it.

sumitarora:
@clbond Ok that should be good then 😄 :shipit:

clbond:
Thank you!

PR #635, opened 2016-09-12
Introduce keyboard navigation for the component tree (behaves the same as the Chrome "Elements" keyboard navigation)

clbond:
Also remove keycode library since Chrome provides this functionality on its own (thank you @xorgy)


sumitarora:
This is great 😄 🚀 :shipit:

PR #636, opened 2016-09-12
Two critical bug fixes: one that prevents huge numbers of messages from clogging the communication pipe and causing latency; and another that causes @Input() decorators to be omitted from state view

clbond:


clbond:
I am merging this because it causes very undesirable behaviour when debugging certain applications and makes the experience of using Augury feel slow and painful. As well, the fix is extremely simple and therefore extremely unlikely to have negative consequences

PR #642, opened 2016-09-12
Resolve uncaught exception

clbond:


src/backend/utils/highlighter.ts
```diff
@@ -66,6 +66,8 @@ export const highlight = (nodes: Array<Node>) => {
     clear();
   }
 
+  nodes = nodes.filter(n => n != null);
```
xorgy: Since this is all in scope, maybe it would be fine to just use it in the loop at line 75.

xorgy:
Fair enough, I would've gone for an `if` statement in the loop, since it's a loop anyway; but this is fine.

:shipit:

Issue #640, opened 2016-09-12
Expand/Collapse Icon do not work


igor-ka:
Steps to reproduce:

1) Go to the demo kitchen sink app
2) Pick a component with State and collapse/expand it
3) Select Counter component in the InputOutput section
4) Go back to the component selected in step 2 above, note that State expansion/collapsing does not respond.

This happens in the latests dev branch.

clbond:
I can't reproduce this. Can you stop by my desk and show me

clbond:
If you open the nested Dev Tools (to debug Augury itself) is there an exception in the console? If so does it say something about highlighting?

igor-ka:
The issue is addressed for me

PR #643, opened 2016-09-12
Resolve uncaught exception that @igor-ka found in @Input() alias template

clbond:
Resolves this issue: https://github.com/rangle/augury/issues/640


xorgy:
:shipit:

I'm embarassed to have committed this bug. ;- )

PR #646, opened 2016-09-12
Say to update popup.html with each version in release-process.md

xorgy:
`popup.html` has the version number in it, so it needs to be updated with the version number when we do a release.


igor-ka:
:shipit:

PR #655, opened 2016-09-13
Move search box into components search affects, and remove it from header.

### LINKED ISSUE #566: Integrate the Find bar with the component tree directly, instead of having it outside.
Currently the Find bar is on the main panel, and needs to be enabled/disabled based on the main view tab state. This does not give the impression that the Find bar belongs to the component tree.

xorgy:
- Add search box to the TreeView component.
 - Add search box to the RouterTree component.
 - Clean out unnecessary search events and values from app-trees, header, and frontend.
 - Fix forward and backward search keybinding logic in the Search component.
 - Make the header less tall.

This resolves #566


sumitarora:
:shipit: 🚀

PR #658, opened 2016-09-14
Dissolve the Header and put the settings button in the toplevel tab bar.

### LINKED ISSUE #656: Dissolve the Header component and move the settings button to the toplevel tab bar.
The Header component is now mostly empty, so it makes sense to simply move what remains of it into the toplevel tab bar.

xorgy:
This resolves #656


sumitarora:
:shipit:

PR #668, opened 2016-09-15
adding logo back in header

### LINKED ISSUE #664: Regression: Add Augury Logo back
Should bring back Augury logo to the top left corner of the view, i.e. left of the Component Tree tab.

sumitarora:
Should resolve #664

image

### DIFF
--- a/src/frontend/components/app-trees/app-trees.html
+++ b/src/frontend/components/app-trees/app-trees.html
@@ -1,4 +1,5 @@
 <bt-tab-menu
+   showLogo="true"
    [tabs]="tabs"
    (tabChange)="onTabSelectionChanged($event)"
    [selectedTab]="selectedTab">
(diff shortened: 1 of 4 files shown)


igor-ka:
Hey @sumitarora looks good overall, the only thing is that i would adjust the padding (or whatever) to position it evenly between the first tab on the right and the dev tools viewport border on the left.
screen shot 2016-09-15 at 10 07 59 am

igor-ka:
@sumitarora also the logo appears in 2 places like in the screenshot below...

screen shot 2016-09-15 at 10 56 34 am

xorgy:
@sumitarora Take a look at how I got the settings button into the top tab bar, should be a similar procedure. (Except you'll want the transclusion slot for the logo to be inside the tab container, rather than outside).

sumitarora:
@xorgy I had to do it using flag `showLogo` due to issue in transclusion that content is rendered after initial content always here is the issue related to that https://github.com/angular/angular/issues/6164

igor-ka:
@sumitarora nice! LGTM :shipit:

PR #671, opened 2016-09-15
Port to Angular 2.0.0

### LINKED ISSUE #660: Support debugging Angular 2.0.0 applications

xorgy:
This resolves #660


example-apps/kitchen-sink-example/source/components/form-controls/control-form.ts
```diff
@@ -1,16 +1,15 @@
...
+} from '@angular/forms';
 
 @Component({
   selector: 'control-form',
-  directives: [REACTIVE_FORM_DIRECTIVES],
-  providers: [FORM_PROVIDERS],
   template: `
     <div>
-        <form (submit)="onSubmit(formControl.value)"
+        <form (submit)="onSubmit(formGroup.value)"
```
xorgy: As a note to reviewer, I think this just fixes an existing bug. There was no `formControl` property on the component before I started, and it works now that it's `formGroup`.

sumitarora:
@xorgy I have reviewed the PR. `Inputs` and `Outputs` are not different now. But there is `listeners` array in the component which holds the subscriptions for the component we can compare both and get outputs array.

xorgy:
@sumitarora Thanks for the insight. I'll finish up tomorrow morning thanks to you. :- )

xorgy:
@sumitarora I've resolved the issue with @Input and @Output and I think this is ready for final review/test.

example-apps/kitchen-sink-example/source/app.routes.ts
```diff
@@ -1,11 +1,11 @@
-import { provideRouter, RouterConfig, RouterModule }  from '@angular/router';
+import { RouterModule }  from '@angular/router';
 
 import { KitchenSinkRoutes, KitchenSinkDeclarations }
  from './containers/kitchen-sink.routes';
 import { RouterRoutes, RouterDeclarations }
  from './components/router/router.routes';
 
-export const routes: RouterConfig = [
+export const routes = [
```
igor-ka: @xorgy is there no more type for this?

src/frontend/components/component-info/component-info.ts
```diff
@@ -32,68 +32,47 @@ export class ComponentInfo {
...
+  ngOnInit() {
+    this.ngOnChanges();
   }
 
-  private get hasProviders() {
-    return this.node &&
-        this.node.providers &&
-        this.node.providers.length > 0;
-  }
+  ngOnChanges() {
```
igor-ka: @xorgy a lot of this stuff has been moved to the onChanges hook, which implies that components can have providers or other things added/removed at runtime.

I am almost certain that I am missing something, and probably out of the loop, but would like to understand the reasons behind this resetting everything everytime onChanges gets called.

sumitarora:
👍 looks good please rebase with dev and :shipit:

example-apps/kitchen-sink-example/source/app.routes.ts
xorgy: Yeah, there seems to be no longer a type for this; or at least it isn't the same. I don't see why they would just change it, so I suspect it was just removed. I could look into that.

src/frontend/components/component-info/component-info.ts
xorgy: @igor-ka Because OnChanges happens less often than a template render (which was what previously triggered these checks. I put them in OnChanges conservatively; I'm not sure if we could just make it OnInit.

example-apps/kitchen-sink-example/source/app.routes.ts
igor-ka: @xorgy no need to change it I was just curious. Not worth looking into it.

sumitarora: @xorgy If you just type it to `Routes` this should work

src/frontend/components/component-info/component-info.ts
xorgy: OnChanges includes changing the node, which changes essentially all of these properties anyway.

Also note that previously these were getters, some of them referenced several times in the template, which means that (although they don't change) the values were rebuilt several times per template render.

igor-ka: @xorgy thank you for clarifying. Sounds good to me.

PR #673, opened 2016-09-16
Update the z-index value of the settings menu to fix overlapping issues.

### LINKED ISSUE #661: Z-index issues with the settings menu.
When the settings menu is opened on the properties pane or the router graph, some elements from "below" the settings menu are still shown. Screenshots attached.

screen shot 2016-09-14 at 11 43 12 am
screen shot 2016-09-14 at 11 43 44 am

hilaryhacksel:
Resolves #661

### DIFF
--- a/src/styles/components/header.css
+++ b/src/styles/components/header.css
@@ -10,6 +10,7 @@
 
 .setting-dropdown {
   -webkit-user-select: none;
+  z-index: 200;
 
   & input {
     vertical-align: text-bottom;


igor-ka:
@hilaryhacksel Thank you. Can you add this as a commit to an existing branch instead? It should update the older PR for this issue automatically. I'd prefer we do it this way, so as to keep track of the comment history.

Let me know if I should clarify :)

PR #675, opened 2016-09-16
Fix the growing and shrinking property inputs

### LINKED ISSUE #652: Property lines in the state editor become taller when the property editor enters edit mode.

hilaryhacksel:
Resolves #652


igor-ka:
@hilaryhacksel I've tested this and the change LGTM visually. 👍

PR #688, opened 2016-09-20
Simplify and minify the production CRX

xorgy:
By removing the sourcemaps and src, and turning on minification for the production build, I have reduced the size of the augury CRX from about 3.3MiB to about 0.8MiB. This also has a positive impact on startup time.


hilaryhacksel:
👍

PR #690, opened 2016-09-20
Remove the minimize webpack optimization.

xorgy:
`--optimize-minimize` interferes with `@Input` an `@Output` detection.


xorgy:
Preapproved by @igor-ka

PR #692, opened 2016-09-21
new superhero app using updated angular cli

hilaryhacksel:


xorgy:
Looks good :shipit:. though I'm not sure why you kept the old one around. ;- )

PR #693, opened 2016-09-21
Update the architecture documentation (I will update the diagrams tomorrow)

clbond:


docs/ARCHITECTURE.md
```diff
@@ -1,45 +1,496 @@
...
+
+```
+0 0 1 2 0 1 4 2 0
+```
+
+This ID doubles as a **path** to that node inside of the `MutableTree`. First,
+recall that a `MutableTree` can contain multiple application component trees.
+So the first number in the path is the index of the application that the node
+belongs to. Each subsequent number is an index into the `children` array inside
+of the parent node. Therefore, if you have an ID, you now how to traverse the
```
igor-ka: know instead of now?

Issue #684, opened 2016-09-19
$a console command sometimes throws an error


hilaryhacksel:
Steps to reproduce:
1. Open a new tab and make sure your developers tools window is closed.
2. Navigate to your angular app (or https://augury.angular.io/demo/#/).
3. Open your developers tools and pick the Augury tab.
5. Click on some Angular component.
6. Type $a into the console and press enter
-> produces the error: Uncaught ReferenceError: $a is not defined(…)

However, $a is defined if your developer tools is open before navigating to the angular app.

xorgy:
Interesting.

orizens:
hi @hilaryhacksel 
I investigated this issue.
It seems like the 'window' referenced within the [consoleReference function](https://github.com/rangle/augury/blob/dev/src/backend/backend.ts#L279) is not the actual window of the relevant tab. 
This is caused of a limitation with content scripts as [described here](https://developer.chrome.com/extensions/content_scripts) 
It seems like the proper solution would be to use postMessage to interact with the Tab's window object and set the '$a' property there.

stevenkampen:
Ok, I am able to reproduce this reliably now, and to do that I had to include the following steps:
1. Make sure you are signed OUT of Chrome.
2. Open the dev tools with "Inspect Element", rather than CMD + Shift + J.

A cache refresh was not required but the quitting chrome and opening again is.

stevenkampen:
@orizens I think I've found at least the superficial cause of this error. At least it worked for me and for @hilaryhacksel. It might make you laugh but in the end it was just that the incorrect console "context" is selected. It should be "top" as in the screenshot below, but I found that mine. Please confirm if this is describes your situation also.

<img width="639" alt="screen shot 2016-10-07 at 4 09 09 pm" src="https://cloud.githubusercontent.com/assets/898763/19195508/d89e54f2-8ca9-11e6-99eb-ee833c73093b.png">

I'm thinking now about how to mitigate this annoyance with some proactive hinting in the console.

PR #710, opened 2016-09-29
issue #684:  console command sometimes throws error

orizens:
This should fix issue #684.
It sends the current correct window object to the consoleReference Function each time.

### DIFF
--- a/src/backend/backend.ts
+++ b/src/backend/backend.ts
@@ -141,7 +141,7 @@ const messageHandler = (message: Message<any>) => {
 
         const node = previousTree.traverse(path);
 
-        this.consoleReference(node);
+        this.consoleReference(node, () => window);
 
         // For component selection events, we respond with component instance
         // properties for the selected node. If we had to serialize the
@@ -273,10 +273,10 @@ export const routerTree = (): Array<MainRoute> => {
   return routes;
 };
 
-export const consoleReference = (node: Node) => {
+export const consoleReference = (node: Node, windowRef: Function) => {
   const propertyKey = '$a';
 
-  Object.defineProperty(window, propertyKey, {
+  Object.defineProperty(windowRef(), propertyKey, {
     get: () => {
       if (node) {
         return ng.probe(node.nativeElement());


sumitarora:
💯 :shipit: @hilaryhacksel ❓

hilaryhacksel:
Hi @orizens, thanks so much for looking into this. I tried testing your fix and I'm still seeing this issue?

orizens:
hi @hilaryhacksel - well, you're right.
perhaps i had caching issues. i'm checking into that.

hilaryhacksel:
Thanks @orizens !

orizens:
@hilaryhacksel 
It seems like there's another **ghost** "instance" of the same window.  
In this screenshot, i added a dummy property "WINDOW_OREN" (as an object) while in "this.consoleReference"  (this is the **ghost** instance of window). This window has the url of augury demo page (augury.io..). 
Later, in the console, when i try to reference the "WINDOW_OREN" in document, it is undefined. 
So - I'm stuck with the question why there are 2 instances of the same window.
screen shot 2016-09-29 at 10 43 12 pm

stevenkampen:
I've not been able to reproduce this repeatably (only 3 or 4 time now out of about 400 tries). However, for others it's far more common, so as long as someone in that position says the fix works for them, then I think we're good. @hilaryhacksel The fix *does* work for you right?

orizens:
@stevenkampen i'm able to reproduce this issue after cleaning the cache, closing the demo page on which i check it and open a new tab with it.
unfortunately, the fix i applied is not the answer. 
As I indicated before, there seem be another instance of a window object which includes the same dom but its not the visible window that this extension needs to operate on.

stevenkampen:
@orizens I can now reproduce this, and I've left the additional steps on the original issue. For me I had to make sure I was signed out of chrome (not just incognito even), and ensure that I'm opening the dev tools with "inspect element" rather than CMD + Shift + J (so the console tab isn't opened by default?). I'm going to use these clues to try to find the next step.

stevenkampen:
Replaced with: #730

PR #722, opened 2016-10-05
Add pared-down version of the kitchen sink demo.

xorgy:
Kitchen Sink includes a number of demonstrations for things which Augury does not help with, the new demo should only include things which Augury can visualize or inspect.

### DIFF
--- a/example-apps/demo/README.md
+++ b/example-apps/demo/README.md
@@ -1,0 +1,8 @@
+# Augury Taster
+
+Use this application to get familiar with Augury
+
+## Building and Serving
+
+1. Build the application by running the `npm install` command.
+2. Serve the application by running the `npm start` command.
--- a/example-apps/demo/.gitignore
+++ b/example-apps/demo/.gitignore
@@ -1,0 +1,2 @@
+node_modules
+dist
--- a/example-apps/demo/package.json
+++ b/example-apps/demo/package.json
@@ -1,0 +1,32 @@
+{
+  "name": "augury-demo",
+  "scripts": {
+    "postinstall": "rimraf typings && typings install",
+    "start": "webpack-dev-server --display-error-details --display-cached",
+    "build": "webpack --display-error-details --display-cached"
+  },
+  "devDependencies": {},
+  "dependencies": {
+    "@angular/common": "2.0.0",
+    "@angular/compiler": "2.0.0",
+    "@angular/core": "2.0.0",
+    "@angular/forms": "2.0.0",
+    "@angular/http": "2.0.0",
+    "@angular/platform-browser": "2.0.0",
+    "@angular/platform-browser-dynamic": "2.0.0",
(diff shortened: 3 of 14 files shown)


example-apps/demo/source/components/injector-tree/injector-tree-dummy-1.ts
```diff
@@ -0,0 +1,26 @@
...
+  selector: 'injector-tree-dummy-1',
+  template: `
+    <p>injector-tree-dummy-1 init: service3</p>
+    {{service3Value}}
+    <hr/>
+    <injector-tree-dummy-3></injector-tree-dummy-3>
+    <injector-tree-dummy-4></injector-tree-dummy-4>
+  `
+})
+export default class InjectorTreeDummy1 {
```
xorgy: Honestly, it didn't cross my mind to change it, but since we've been consistent in Augury itself I'll change it here as well, thanks.

example-apps/demo/source/components/router-toy/data-2.ts
```diff
@@ -0,0 +1,31 @@
...
+})
+export default class RouterToyData2 {
+  public message: string;
+  public name: string;
+  private params$: any;
+
+  constructor(private activatedRoute: ActivatedRoute) { }
+
+  ngOnInit() {
+    this.params$ = this.activatedRoute.params.subscribe(({name, message}) => {
```
xorgy: Maybe, but for now we don't really have any functionality which inspects that, this is basically just here so that the router links don't look stupid.

example-apps/demo/source/components/injector-tree/injector-tree-dummy-1.ts
igor-ka: No biggie, but thanks!

example-apps/demo/source/components/router-toy/data-2.ts
igor-ka: ahh got it thanks!

example-apps/demo/README.md
```diff
@@ -0,0 +1,8 @@
+# Augury Taster
```
sumitarora: `Taster` Is the name correct?

xorgy: Fourth sense here.
https://en.wiktionary.org/wiki/taster

xorgy:
For what it's worth, I don't think there's any rush. It's not like we lose anything by having this sit in the branch until it's basically perfect. And even if we want to do PRs, we can just PR into new-demo instead of dev.

xorgy:
It would make things more convenient to merge this now I think. Let me know if you folks agree.

xorgy:
@stevenkampen @sumitarora @igor-ka 
It would make things more convenient to merge this now I think. Let me know if you folks agree.

Issue #674, opened 2016-09-16
When a component is selected, one can not scroll it out of view.


xorgy:
When a component is selected, and one tries to scroll it out of the view, any change detection cycle will cause the view to jump back to the selected element.

PR #724, opened 2016-10-06
Resolve bizarre scrolling issue by removing the "auto scroll" code

clbond:
See attached video for what I mean
[scrolling problem.zip](https://github.com/rangle/augury/files/514162/scrolling.problem.zip)

### DIFF
--- a/src/frontend/components/component-tree/component-tree.ts
+++ b/src/frontend/components/component-tree/component-tree.ts
@@ -39,30 +39,6 @@ export class ComponentTree {
     private el: ElementRef
   ) {}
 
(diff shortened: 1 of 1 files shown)


xorgy:
@clbond https://github.com/rangle/augury/issues/674 this is the issue report for this bug.

xorgy:
I think maybe we should find some way to make it scroll into view when the user explicitly selects (through search or up/down arrows), but not at random view check points.

xorgy:
I've superseded this with #725 which solves the issue while maintaining the functionality.

PR #730, opened 2016-10-11
Add execution context warning to content script

### LINKED ISSUE #684: $a console command sometimes throws an error

stevenkampen:
resolves #684 

@xorgy I spent some time looking at the chrome APIs to see how we might be able to add this warning to all execution contexts except the "top" one, but I didn't find anything conclusive (though I did read that it *should* be possible for each execution context to access the `window` object of each of the others).

This adds it in the parent extension context only. If you have any suggestion for improvements, please point me in the right direction.

### DIFF
--- a/src/backend/backend.ts
+++ b/src/backend/backend.ts
@@ -121,6 +121,12 @@ checkDebug(() => {
   getAllAngularRootElements().forEach(root => bind(ng.probe(root)));
 });
 
+const selectedComponentPropertyKey = '$a';
+const noSelectedComponentWarningText = 'There is no component selected.';
+
+Object.defineProperty(window, selectedComponentPropertyKey,
+  { value: noSelectedComponentWarningText });
+
 const messageHandler = (message: Message<any>) => {
   switch (message.messageType) {
     case MessageType.Initialize:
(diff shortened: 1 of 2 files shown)


stevenkampen:
Solves #684

stevenkampen:
@igor-ka It's simple enough to define `$a` (in the top context) as a warning string initially ("no component selected"), then as soon as something is selected that would go away. The same thought occurred to me while looking at this. I think that's a safe/small enough change that it can go in with this. I'll add it now.

src/backend/backend.ts
```diff
@@ -121,6 +121,12 @@ checkDebug(() => {
   getAllAngularRootElements().forEach(root => bind(ng.probe(root)));
 });
 
+const selectedComponentPropertyKey = '$a';
+const noSelectedComponentWarningText = 'There is no component selected.';
```
stevenkampen: @igor-ka This shows now when no component is selected.

PR #738, opened 2016-10-17
Update router tree with changes

stevenkampen:
@igor-ka @xorgy @clbond This isn't complete yet, but I wanted to get some eyes on it. These changes handle clearing and re-rendering the router tree on data change, as well as having the routes loaded from the point as the component tree, rather than on tab change (to support router tree updates while that tab is open).

The one thing I saw was some intermittent issues (not rendered properly) when multiple `RouterTree` messages are sent initially. I'm looking further into that, but if anyone has any thoughts, let me know. As a soft workaround I make sure it also triggers an `Input` change on tab switching, so switching away and back will give it another try.

The remaining work required is to support partial router tree updates as the component tree does, which I'm taking a stab at now.

### DIFF
--- a/example-apps/kitchen-sink-example/source/components/lazy-load/lazy-load.ts
+++ b/example-apps/kitchen-sink-example/source/components/lazy-load/lazy-load.ts
@@ -1,0 +1,19 @@
+import {Component} from '@angular/core';
+import { NgModule }           from '@angular/core';
+import { Routes, RouterModule }  from '@angular/router';
+
+
+@Component({
+  template: `<H1>Lazy Loaded Component One</H1>`
+})
+class LazyLoadedComponentOne { }
+
+const routes: Routes = [
+  { path: 'lazy1', component: LazyLoadedComponentOne },
+];
+
+@NgModule({
+  imports: [  RouterModule.forChild(routes) ],
+  declarations: [ LazyLoadedComponentOne ]
+})
+export class LazyLoadedModule { }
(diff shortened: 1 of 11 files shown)


igor-ka:
@stevenkampen so far looks good to me overall.

I see that we do not diff the router tree (like we do with the component tree). I am assuming that it is not necessary since it's smaller and does not change as often as component tree? I just want to clarify, and could definitely be missing something.

@clbond could you have a look at this one as well once you have a few spare minutes?

stevenkampen:
@igor-ka All good now?

example-apps/kitchen-sink-example/source/components/lazy-load/lazy-load.ts
```diff
@@ -0,0 +1,19 @@
+import {Component} from '@angular/core';
```
igor-ka: Minor detail but can you add spaces @stevenkampen for consistency :)

src/backend/utils/parse-router.ts
```diff
@@ -50,5 +50,11 @@ function assignChildrenToParent(parent, children): [any] {
 }
 
 function childRouteName(child): string {
-  return child.component ? child.component.name : 'no-name-route';
+  if (child.component) {
```
igor-ka: i'd prefer the if/elseif/else here. Or am I missing something?

stevenkampen: @igor-ka My thinking was that because they are returns, it would be a little redundant, but that's not a strong opinion. Change made.

stevenkampen: @igor-ka Absolutely.

Issue #734, opened 2016-10-15
Error when logon asana.com website with extension enable


fpragana:
I trying ope the website www.asana.com but when i make the logon the Chrome throw a JS erro. When i disable the augury extension, the problem is solved. 

<img width="1336" alt="screenshot 2016-10-15 16 29 00" src="https://cloud.githubusercontent.com/assets/3301373/19412835/a7002140-92f4-11e6-96e9-976437d6b5fa.png">
<img width="1328" alt="screenshot 2016-10-15 16 29 27" src="https://cloud.githubusercontent.com/assets/3301373/19412836/a70081c6-92f4-11e6-9961-8133ea1c99b0.png">

PR #739, opened 2016-10-17
Resolve issue with KeyRegistry blowing up on applications with weird ES6 shims

clbond:
This PR resolves issues with sites like Asana.com which include weird ES6 shim implementations. Essentially this change prevents inclusion of Angular2 code before the ng-validate gate has been passed. (Prior to this change, Angular2 would be included as part of the dependency tree of ng-validate and caused exceptions on sites that use nonstandard ES6 shim implementations.)

### DIFF
--- a/src/communication/message-factory.ts
+++ b/src/communication/message-factory.ts
@@ -1,5 +1,3 @@
-import {DebugElement} from '@angular/core';
-
 import {
   Message,
   MessageResponse,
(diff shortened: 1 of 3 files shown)


igor-ka:
This PR should resolve #734

stevenkampen:
@igor-ka @clbond  Oh I had two versions of the extension installed. Works now. :shipit:

PR #741, opened 2016-10-18
Fix "application error" communication, introduce new class of error for AoT applications, introduce performance optimization for applications that are not being debugged

clbond:
* Resolve problem with ApplicationError communication between backend and frontend
* Add new class of error for applications built with AoT enabled and cannot be debugged
* Introduce new performance optimization where we do not subscribe to Angular changes unless we know the developer has actually opened the wAugury tab

### DIFF
--- a/src/backend/backend.ts
+++ b/src/backend/backend.ts
@@ -1,6 +1,9 @@
 import {DebugElement} from '@angular/core';
 
-import {Subject} from 'rxjs';
+import {
+  Subject,
+  Subscription,
+} from 'rxjs';
 
 import {
   Metadata,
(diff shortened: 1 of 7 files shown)


igor-ka:
everything looks good @clbond, just a couple of questions in my comments. Other than that I think the best would be to describe the story for this under a separate issue so as to know what to expect this code to do. What do you think?

src/backend/backend.ts
```diff
@@ -92,13 +95,25 @@ const update = () => {
 
 const subject = new Subject<void>();
 
+const subscriptions = new Array<Subscription>();
+
 const bind = (root: DebugElement) => {
+  if (root.injector == null) {
+    // In AoT-enabled debug builds, the data we get back from {@link ng.probe}
```
tbosch: This is not completely true. A user has to do both:
1) use AOT in debug mode and
2) NOT call `enableProdMode()`

Then this data is filled, even with AOT.

clbond:
This resolves https://github.com/rangle/augury/issues/742

clbond:
I removed the comment and made the error messages more generic.

PR #744, opened 2016-10-19
Decorators revamp part 1

xorgy:
This is not totally finished, it turns out there is some work required to support certain groups of decorators properly. The changes in this PR fix detection of the currently supported `@Input` and `@Output` decorators which were previously incorrect. I will publish a separate PR which expands to support multiple decorators on a single member, and all of the core Angular Decorators.

This PR also includes a switch to ES6 codegen in typescript, which works perfectly fine in Chromes released in the last 12 months or so, and makes it much easier to debug the generated code since there is less mangling.

In addition, I have also removed some properties on Node which are no longer used, and the code that attaches them.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,8 +60,6 @@
     "basscss-layout": "^3.1.0",
     "clone": "^1.0.2",
     "css-loader": "^0.21.0",
-    "es6-promise": "^3.1.2",
-    "es6-shim": "^0.35.0",
     "file-loader": "^0.8.5",
     "msgpack-lite": "^0.1.20",
     "object-assign": "4.0.1",
--- a/src/backend/backend.ts
+++ b/src/backend/backend.ts
@@ -224,10 +224,7 @@ const getComponentInstance = (tree: MutableTree, node: Node) => {
   if (node) {
     const probed = ng.probe(node.nativeElement());
     if (probed) {
-      return instanceWithMetadata(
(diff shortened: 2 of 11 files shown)


stevenkampen:
@xorgy Tell me if I'm doing something wrong, but `npm run build` (after fresh npm install) now gives me this error:
```
ERROR in ./src/communication/message-factory.ts
(54,12): error TS2304: Cannot find name 'create'.
```

xorgy:
@stevenkampen Indeed, it seems like somebody added another message type after my base branch, and it didn't show up as a conflict since it was just inserts.

stevenkampen:
@xorgy This builds fine now, but during a smoke test I'm seeing an issue with the state info panel, or whatever we call it. Expanding an Input that has an Object value results in a blank space where the properties would usually be, and the arrow becomes unresponsive.

screen shot 2016-10-21 at 3 29 28 pm

xorgy:
@stevenkampen Interesting, I don't see how that could have changed, are you sure that's a regression?

xorgy:
@stevenkampen thanks for looking into it. This is going to kill me. 
:- \

igor-ka:
@stevenkampen @clbond I am closing this PR. @stevenkampen I remember that you mentioned that there is a code in there that contains a bug fix for some issue, I want to make sure we don't loose track if it is still needed.

PR #746, opened 2016-10-21
On push property change detect

### LINKED ISSUE #559: The component state editor does not trigger change detection for OnPush components.
In the cases where a component with a change detection strategy of ON_PUSH is selected, and the state of that component is modified on the Properties Pane, the UI is not updated unless change detection is triggered by some other means.

Should somehow trigger change detection programatically, in the cases where the state for said component is modified on the Properties Pane.

stevenkampen:
resolves #559 

This calls `markForCheck()` and `checkForUpdates()` for all properties on an `ChangeDetection.OnPush` component, when a property is updated, as well as fixes an issue with resolving the ChangeDetectionStrategy (when it's `OnPush`).

### DIFF
--- a/src/backend/backend.ts
+++ b/src/backend/backend.ts
@@ -253,6 +253,12 @@ const updateProperty = (tree: MutableTree, path: Path, newValue) => {
       const instanceParent = getNodeInstanceParent(probed, path);
       if (instanceParent) {
         instanceParent[path[path.length - 1]] = newValue;
+        if (node.changeDetection === 'OnPush') {
+          probed.childNodes.map(childNode => {
+            childNode._debugInfo._view.changeDetectorRef.markForCheck();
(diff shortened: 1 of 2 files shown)


stevenkampen:
@igor-ka said :shipit:, but github being down prevented him from saying it himself. I swear.

src/backend/backend.ts
```diff
@@ -253,6 +253,12 @@ const updateProperty = (tree: MutableTree, path: Path, newValue) => {
       const instanceParent = getNodeInstanceParent(probed, path);
       if (instanceParent) {
         instanceParent[path[path.length - 1]] = newValue;
+        if (node.changeDetection === 'OnPush') {
+          probed.childNodes.map(childNode => {
+            childNode._debugInfo._view.changeDetectorRef.markForCheck();
```
igor-ka: @ericjim do you have any insight on @stevenkampen's comment by any chance?

PR #748, opened 2016-10-21
Resolve performance issue due to bug in which JSON patch library is not able to compare `undefined` values

clbond:
* Resolve bug in which JSON patch library is not able to compare `undefined` values properly.  In turn, this resolves a major performance issue where Augury was generating endless 'changes' even when nothing changed, if there was an `undefined` value somewhere in the tree structure.
* Add unit tests to cover this case and verify that the patch library continues to work as expected for non-undefined values

### DIFF
--- a/src/utils/patch.test.ts
+++ b/src/utils/patch.test.ts
@@ -1,0 +1,56 @@
+import * as test from 'tape';
+
+import {compare} from './patch';
+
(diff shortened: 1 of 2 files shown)


stevenkampen:
@clbond LGTM. :shipit:

PR #751, opened 2016-10-25
Make sure emitter values are emitted within the ngZone

### LINKED ISSUE #752: Events emitted from Augury don't result in component tree updates when they should
I suspect this is ngZone related (I know it it because I already fixed it! shhh!)

stevenkampen:
This resolves an issue where changes to an inspected app that are a result of events emitted from Augury (as opposed to the same events emitted naturally through the app), do not result in the component tree being updated.

resolves #752

@vanessayuenn FYI.


stevenkampen:
resolves #752

stevenkampen:
@igor-ka Please approve. Release incoming.

stevenkampen:
Will do.

PR #769, opened 2016-11-07
Fix multiple applications on same page (was just a styling problem)

clbond:
Prior to this patch, if there were multiple Angular 2 apps on the same page, they would not be displayed due to a styling error.


clbond:
Attached is an Angular 2 test app that has multiple roots on the same page. Untar it and do `npm install` then `ng serve`.
[multi-app.zip](https://github.com/rangle/augury/files/576451/multi-app.zip)

stevenkampen:
@clbond LGTM :shipit:

PR #770, opened 2016-11-08
Add support for @ViewChild, @ViewChildren, @ContentChild and @ContentChildren decorators in state view

clbond:
* Add support for @ViewChild, @ViewChildren, @ContentChild and @ContentChildren decorators in state view
* Refactor all 'inputs' and 'outputs' to use property metadata system so that they continue to work at all levels of the state tree, not just the toplevel (this is pertinent when using eg @ViewChild and the child component has @Input()s)

Screenshot:
screen shot 2016-11-07 at 7 32 50 pm

### DIFF
--- a/src/frontend/components/component-info/component-info.ts
+++ b/src/frontend/components/component-info/component-info.ts
@@ -35,7 +35,7 @@ export class ComponentInfo {
 
   private path: Path;
 
-  ngOnChanges(changes: SimpleChanges) {
+  ngOnChanges() {
     if (this.node) {
       this.path = deserializePath(this.node.id);
     }
--- a/src/frontend/components/component-info/component-info.ts
+++ b/src/frontend/components/component-info/component-info.ts
@@ -3,10 +3,11 @@ import {
   EventEmitter,
   Input,
   Output,
+  SimpleChanges,
 } from '@angular/core';
 
 import {ComponentLoadState} from '../../state';
-import {InputOutput} from '../../../frontend/utils';
+
 import {UserActions} from '../../actions/user-actions/user-actions';
 
 import {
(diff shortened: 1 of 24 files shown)


src/frontend/components/component-info/component-info.ts
```diff
@@ -22,44 +23,21 @@ import {
...
-  inputs: InputOutput;
-  outputs: InputOutput;
-
-  ngOnInit() {
-    this.ngOnChanges();
-  }
+  private path: Path;
 
-  ngOnChanges() {
+  ngOnChanges(changes: SimpleChanges) {
```
igor-ka: @clbond I don't see changes being used. Is it added just to conform with the signature of the method?

src/tree/decorators.ts
```diff
@@ -0,0 +1,90 @@
+import {
+  Component,
+  DebugElement,
+} from '@angular/core';
```
igor-ka: @clbond is this used on the backend? If so I'd like us to avoid importing anything from angular, or any other libraries.

src/frontend/components/component-info/component-info.ts
clbond: Good point, fixed

src/tree/decorators.ts
clbond: These imports were already in place, they were just in a separate file, `transform.ts`. I am not sure how to implement this code without at least `Component`.

igor-ka: @clbond if it was already there we can remove it under a different issue.
In this case we can either us creating our own type for this, or just use `any`.

clbond: Alright, fair enough

Issue #775, opened 2016-11-08
Display real type information in RenderState instead of just "Object"


clbond:
In the State panel, we currently display most objects like:

`foo: Object`

Where it would be better to display the real type information, like:

`foo: MyComponent`

PR #776, opened 2016-11-08
Display real type information in the State view instead of always "Object"

clbond:
Now we retain the name of the type across backend->frontend serialization so that we can display the real object type in the "State" view. See:
screen shot 2016-11-08 at 12 25 30 pm

Resolves: https://github.com/rangle/augury/issues/775

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,6 @@
     "es6-promise": "^3.1.2",
     "es6-shim": "^0.35.0",
     "file-loader": "^0.8.5",
-    "md5": "^2.2.1",
     "msgpack-lite": "^0.1.20",
     "object-assign": "4.0.1",
     "postcss-cssnext": "^2.5.2",
--- a/src/frontend/components/component-info/component-info.html
+++ b/src/frontend/components/component-info/component-info.html
@@ -68,6 +68,7 @@
             <template [ngSwitchCase]="ComponentLoadState.Received">
               <bt-render-state
                  [id]="node.id"
+                 [componentMetadata]="componentMetadata"
                  [metadata]="metadata"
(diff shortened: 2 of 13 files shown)


igor-ka:
@clbond I am assuming we do not display primitives types like string or number because there is no constructor property?

clbond:
@igor-ka We display primitive types using `bt-property-editor`, so their types would never be shown on the UI. We let the user edit primitive values.

igor-ka:
@clbond yes i understand, I am wondering if it makes sense to display both the (primitive) type as well as let users edit them. It seems like an overkill at first thought but i am wondering what you think about this?

clbond:
@igor-ka I had to add a bit of code to this PR because the way I did metadata in the `@ViewChildren` PR was wrong. I was creating a map of Path -> Metadata, but of course an object can appear at multiple paths, in which case the metadata was wrong. So I removed the md5 package and all the path hashing code and replaced it with object references. Our serializer maintains object references across serialization (if the same object appears twice, it will use the same instance), so now the metadata system relies on object references instead of paths when determining what decorators to apply to an object. Sorry if this is confusing ... it's a confusing problem

clbond:
Don't merge this yet

clbond:
OK, I fixed the case I discussed above. Paths are no longer used as keys for the metadata structure. Additionally, we can't store object values as component metadata keys because oftentimes those values are scalar and thus not unique. So we use the component instance itself as the map key and then store a list of properties and their metadata and associate those with the component. This works across all use cases, across circular references, duplicate references, deep nesting of component references, etc. I am confident in this fix.

PR #781, opened 2016-11-11
Add some missing rxjs types to the metadata system

clbond:
There were several types of `Subject`s and `Subscriber`s that we missed in our metadata lookup. Add them.


mobinni:
LGTM 🚀

PR #816, opened 2016-11-23
Feature/662 provider info

### LINKED ISSUE #662: Show all the internal information in Providers accordion
Instead of handpicking which information to show for each of the built in providers as well as ignoring certain providers (like custom directives), show all of the information pertaining to the provider tokens. 

For example:
1) In the Kitchen Sink Demo app click Form Component section
2) Select one of the `input` elements on the section of the component tree that appears
3) Expand the Providers accordion and note the information shown there.
4) Go to the console and input the following `$a.injector.get($a.providerTokens[3])`, this will show the information that corresponds to the `NgControlName` provider.
5) Compare the information on the console with the information under the Providers accordion for the same provider.

We should shall ALL the information for each of the providers, i.e. should show all the same information as is shown on the console.

mobinni:
Resolves #662 

- Added non-component based provider info: ProviderTokens injector data is now showed in the component information panel when an element is selected
- Added library that's able to show JSON to show provider token info - Fix to library currently in PR (https://github.com/mohsen1/json-formatter-js/pull/36) pointing to local forked branch for the fix 
- Fixed color issues select element + added styles for select-element


stevenkampen:
@mobinni I'm seeing some unexpected behaviour in the UI. Using the KitchenSink app, inside the "Form Component" section, I'm seeing different results under "Instance Provider" for the `input` elements, depending on what was selected previously. Selecting the `input` directly after selecting the `form` above it results in no change to instance providers, but selecting the input after another element (an `a` or `li` element above or below it) refreshes it with a different (larger) set of options.

screen shot 2016-11-29 at 3 45 49 pm
screen shot 2016-11-29 at 3 45 28 pm

stevenkampen:
@mobinni I just found that the accordion arrows aren't expanding now.

screen shot 2016-11-30 at 2 38 17 pm

stevenkampen:
@mobinni LGTM. :shipit:

src/frontend/components/component-info/component-info.html
```diff
@@ -4,29 +4,29 @@ <h4 class="primary-color col pt1 pb2 m0">
...
-      *ngIf="node">
+          *ngIf="node">
       ($a in Console)
     </span>
   </div>
 
   <div
-     *ngIf="node && node.changeDetection"
-     class="section-title bg-darken-1 bg-panel primary-color pointer border-bottom py2 px3">
-    <span class="info-key">Change Detection: </span>{{changeDetectionStrategies[node.changeDetection]}}
```
stevenkampen: @mobinni Oh yeah thanks. I forgot to double check that.

src/frontend/components/component-info/component-info.html
```diff
@@ -29,6 +29,7 @@ <h4 class="primary-color col pt1 pb2 m0">
             *ngIf="node && node.description">
           <bt-property-value
             *ngFor="let property of node.description"
+
```
stevenkampen: @mobinni What dis for?? :)

mobinni: woooops

PR #822, opened 2016-11-24
Fix router tree

### LINKED ISSUE #823: Fix broken router tree

mobinni:
fixes router tree and retrieves the outer svg element now from the elementref instead of document.
We were using document.getElementByTagName and the 0'ist tag element on the document. However I added a new svg for select element, so it was referencing the incorrect svg. Now it queryselects from the elementreference of the component instead. Which I assume will not trigger a break anymore.

resolves #823

### DIFF
--- a/src/styles/utils/colors.css
+++ b/src/styles/utils/colors.css
@@ -87,7 +87,7 @@ bt-injector-tree {
     background-color: #E5E5E5;
   }
 
-  & .select-element-active > svg  * > polygon {
+  & .select-element-active > svg * > polygon {
      fill: #5A5A5A;
   }
 
@@ -232,7 +232,7 @@ bt-injector-tree {
     background-color: #313131;
   }
 
-  & .select-element-active > svg  * > polygon {
+  & .select-element-active > svg * > polygon {
     fill: #ccc;
   }
 
--- a/src/styles/utils/colors.css
+++ b/src/styles/utils/colors.css
@@ -87,19 +87,19 @@ bt-injector-tree {
     background-color: #E5E5E5;
   }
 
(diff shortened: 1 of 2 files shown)


src/styles/utils/colors.css
```diff
@@ -87,19 +87,19 @@ bt-injector-tree {
     background-color: #E5E5E5;
   }
 
-  & .select-element-active > svg > * > * > * > polygon {
+  & .select-element-active > svg  * > polygon {
```
stevenkampen: @mobinni No biggie but there's an extra space here.

src/styles/utils/colors.css
```diff
@@ -232,19 +232,24 @@ bt-injector-tree {
     background-color: #313131;
   }
 
-  & .select-element-active > svg > * > * > * > polygon {
+  & .select-element-active > svg  * > polygon {
```
stevenkampen: Also here.

src/styles/utils/colors.css
```diff
@@ -232,19 +232,24 @@ bt-injector-tree {
...
-  & .select-element-active > svg > * > * > * > path {
+  & .select-element-active > svg * > path {
     stroke: #ccc;
   }
 
-  & .select-element-active > svg > * > * > * > polygon {
+  & .select-element-active > svg * > polygon {
     fill: #4280EC;
   }
+<<<<<<< Updated upstream
```
stevenkampen: @mobinni Oops!

src/styles/utils/colors.css
mobinni: Thanks 👍 should be removed now

stevenkampen:
@mobinni LGTM :shipit:

PR #828, opened 2016-11-25
Fix dependency tree support skipself

### LINKED ISSUE #802: Injector Graph wrong service reference arrows
When a service has been instantiated/provided by a component as well as the module or higher level component in the hierarchy, the reference arrows point to an instance of the service from the higher level as opposed to the same level/component in the hierarchy.

stevenkampen:
resolves #802, and adds support for `@SkipSelf` injected parameter decorator

### DIFF
--- a/src/backend/utils/description.ts
+++ b/src/backend/utils/description.ts
@@ -1,5 +1,10 @@
 import {DebugElement} from '@angular/core';
 
+export interface Dependency {
+  type: string;
+  decorators: Array<string>;
+}
+
 export interface Property {
   key: string;
   value;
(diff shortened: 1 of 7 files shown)


mobinni:
what's the best way to test this?

igor-ka:
@mobinni @rajinder-yadav once we decided the best use case to test this, we should probably create an issue to add this into our example app as well as the guide on this feature of augury.

stevenkampen:
@mobinni I've rebased this and tested it. If you could give it another quick smoke test, that would be greeeat.

mobinni:
LGTM 🐸

PR #847, opened 2016-12-01
Task 804 input output example

### LINKED ISSUE #804: Kitchen Sink, Input-Output example, breakout into standalone app.
* Create new stand alone app, move example code from Kitchen Sink.
* Clean up code, make sure to follow ng2 guidelines.
* Create a gif to demo feature in Augury.
* Create a write up guide.
* Work out on any layout design stuff with Fatima.
(*) This will be the blueprint for other stand along app.

rajinder-yadav:
resolved #804


stevenkampen:
@rajinder-yadav Tested working, and I think it's a good example of the input/output feature, but I wonder if the UI could be improved so it's easier to see what you're looking at, at a quick glance. What do you think?

sumitarora:
@rajinder-yadav I agree with @stevenkampen UI needs more ❤️ also few more things.
- Toggle didn't work for me when I manually changed it in augury from `true` to `false`
- It only seems to be working first time might be an augury issue
- I see basscss errors in console while running app
- We should also update readme with description about the example

rajinder-yadav:
@sumitarora in the IO App, state property seems broken is Augury, you're correct it only works once. I also notice the value does not update in Augury if I click on the toggle button.

As for basscss error it looks like it's something to do with the CDN. I'm going to update the code to use their npm module for everything is a plugin!!!

Good idea about adding a link to the guide in the README, might be a better idea to just add a link on the app? Something like, "Take me to the Guide."

rajinder-yadav:
Switched code to use basscss npm modules.

PR #850, opened 2016-12-02
Check router type name instead of just property existence

### LINKED ISSUE #848: [routerTree] Cannot read property 'name' of undefined
I get the following error for an app that doesn't use `angular/router`
```js
Cannot read property 'name' of undefined
```
selection_042

stevenkampen:
resolves #848 

We were not verifying that the injected `router` property was anything in particular, so anything else injected with that property key would break the router parsing.


mobinni:
LGTM 🚀

PR #899, opened 2016-12-28
Feature/ng module context in injector graph, plus NgModules tab

### LINKED ISSUE #764: Add another tab representing the NgModule structure

### LINKED ISSUE #803: Injector Graph should showcase the NgModule associated with the component

stevenkampen:
screen shot 2016-12-28 at 4 23 22 pm

@rajinder-yadav Here is a screenshot of the NgModules tab. If you get a chance, maybe you could test out the feature (using the circle-ci crx) and give me your initial thoughts on it. Cheers!

resolves #764
resolves #803

### DIFF
--- a/src/backend/utils/parse-modules.ts
+++ b/src/backend/utils/parse-modules.ts
@@ -50,7 +50,7 @@ const buildModuleDescription = (module, config) => {
     exports: flatten(config.exports || []).map(ex => parseModuleName(ex)),
     declarations: flatten(config.declarations || []).map(d => d.name),
     providers: flatten(config.providers || []).map(
-      p => typeof p === 'object' && p.provide ? p.provide.name || p.provide.toString() : p.name),
+      p => typeof p === 'object' && p.provide ? p.provide.name || p.provide.toString().replace(' ', ':') : p.name),
   };
 };
 
--- a/src/backend/utils/parse-modules.ts
+++ b/src/backend/utils/parse-modules.ts
@@ -16,7 +16,7 @@ const resolveNgModuleDecoratorConfig = (m) => {
 
 export const parseModules = (firstRootDebugElement: any): {[key: string]: any} => {
   const bootstrappedModule = firstRootDebugElement.injector.get(ng.coreTokens.ApplicationRef)._injector.instance;
-  const [modules, moduleNames] = _parseModule(bootstrappedModule.constructor);
+  const [modules, moduleNames, tokenIdMap] = _parseModule(bootstrappedModule.constructor);
   const serializableModules = {};
   for (const m in modules) {
     if (modules.hasOwnProperty(m)) {
@@ -26,6 +26,7 @@ export const parseModules = (firstRootDebugElement: any): {[key: string]: any} =
   return {
     names: moduleNames,
     configs: serializableModules,
+    tokenIdMap,
   };
 };
 
@@ -41,6 +42,7 @@ const resolveTokenIdMetaData = (token) => {
   if (!Reflect.getMetadata(AUGURY_TOKEN_ID_METADATA_KEY, token)) {
     Reflect.defineMetadata(AUGURY_TOKEN_ID_METADATA_KEY, randomId(), token);
   }
+  return { token: token, augury_token_id: Reflect.getMetadata(AUGURY_TOKEN_ID_METADATA_KEY, token) };
 };
 
 const parseProviderName = p =>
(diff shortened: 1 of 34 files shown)


src/frontend/components/injector-tree/injector-tree.html
```diff
@@ -4,34 +4,41 @@ <h4 class="primary-color pt1 pb2 m0" [hidden]="selectedNode">
       No component selected
     </h4>
   </div>
-
-  <div class="flex-column monospace pl4 pt2"
+  <div class="flex-column monospace pt2"
     [hidden]="!selectedNode"
     [ngClass]="{flex: selectedNode}">
-    <div *ngIf="parentHierarchyDisplay && parentHierarchyDisplay.length > 0">
```
rajinder-yadav: You can use the safe navigation operator `?.` to replace the `&&` shortcut:

This will suffice:
```js
parentHierarchyDisplay?.length > 0
```

stevenkampen:
@rajinder-yadav I had opened this new PR to include both features (#803 and #764), but I didn't see your comments that you mentioned. Is it possible it's still in an "unsubmitted" review? At any rate, this is the new one now, and it has a (ugly) implementation for the module origin feedback in the injector graph. There is lots of supporting code to review though, so I thought we'd get started and I/we can improve the UI over the next little while.

rajinder-yadav:
If you clicked on "Files changed" (on the top) it should show you all my review comments.

stevenkampen:
@rajinder-yadav I do not see them currently.

rajinder-yadav:
I think they are all showing up on this main page, are you seeing those?

stevenkampen:
@rajinder-yadav No I'm not. I believe you still need to "submit" the review. Until then, you are the only one seeing them.

src/frontend/components/injector-tree/injector-tree.ts
```diff
@@ -62,6 +68,8 @@ export class InjectorTree implements OnChanges {
 
     this.parentHierarchy = [{ name: 'root', dependencies: [] }].concat(mainHierarchy).concat([this.selectedNode]);
 
+    this.breadcrumbs = this.parentHierarchy.slice(1, this.parentHierarchy.length);
```
stevenkampen: @rajinder-yadav In this case I do need two separate versions of the array. Just one without the first ("root") element. Removing the second argument to `slice` is good advice though.

src/frontend/components/injector-tree/injector-tree.ts
```diff
@@ -93,6 +102,8 @@ export class InjectorTree implements OnChanges {
 
     let posX, posY, x1, y1, x2, y2;
 
+    const nodesToDraw = [];
```
stevenkampen: @rajinder-yadav Not sure I agree here. To me, using let would suggest to that the variable is expected to be reassigned at some point in this block. My thinking is that devs should understand that not being able to reassign the variable of course does not mean the assigned value can't be mutated. Your way would suggest that const is only used for assignment of immutable values?

src/backend/utils/parse-modules.ts
```diff
@@ -0,0 +1,141 @@
...
+    tokenIdMap,
+  };
+};
+
+const parseModuleName = (m) => {
+  return m.ngModule ? m.ngModule.name : m.name || (m.constructor ? m.constructor.name : null);
+};
+
+const randomId = () => {
+  return Math.random().toString(36).substring(7);
```
stevenkampen: @rajinder-yadav I like the idea of prefixing the `Date.now()`.

stevenkampen: @rajinder-yadav Actually since these are all generated at the same time, I don't think using the date will help much. I have a map of all the ids though, so I've added some code to regenerate on pre-existence.

stevenkampen:
@clbond Would you be able to give this a quick once over also? Just from an architectural point of view.

rajinder-yadav:
LGTM

src/backend/utils/parse-modules.ts
```diff
@@ -0,0 +1,153 @@
...
+    return m.decorators.reduce((prev, curr, idx, decorators) =>
+      prev ? prev : decorators[idx].type.prototype.toString() === '@NgModule' ? decorators[idx].args[0] : null, null);
+  }
+
+  return Reflect.getMetadata('annotations', m).find(decorator => decorator.toString() === '@NgModule');
+};
+
+export const parseModules = (firstRootDebugElement: any): {[key: string]: any} => {
+  // TODO(steven.kampen): This uses a private API. Can it be improved?
+  const bootstrappedModule = firstRootDebugElement.injector.get(ng.coreTokens.ApplicationRef)._injector.instance;
```
clbond: This seems dangerous

stevenkampen: @clbond I've been talking to Igor from the core team about it, and there may be a safer way to do this but I'm waiting on full details.

clbond:
LGTM

PR #901, opened 2016-12-30
Refactor the rendering of providers so that they match the rest of Augury, and make them editable also

clbond:
Instance providers are rendered with the preexisting render-state component now instead of `JSONFormatter`. Removed the `JSONFormatter` dependency. Instance providers are also now editable in the same way that component state is editable. In essence this changeset just makes the Instance Providers panel behave like the rest of Augury.

Screenshot:
screen shot 2016-12-30 at 12 10 12 pm


rajinder-yadav:
Reviewed, looks good.

mobinni:
LGTM

PR #902, opened 2016-12-30
Split apart the frontend NgModule definition and the App component and fix issue with keypresses being swallowed

clbond:
Split apart the frontend NgModule definition and the App component: there is no reason for them to be in the same file

Resolve issue with keypresses being swallowed (thus preventing Chrome shortcuts like Cmd+I from working inside of Augury)


mobinni:
LGTM 🚀 but maybe have someone on the augury team still review it aswell. Nice addition :) @rajinder-yadav

stevenkampen:
@clbond LGTM :shipit:

PR #910, opened 2017-01-11
Feature/auto router resolution

### LINKED ISSUE #829: Support router registration a la https://github.com/angular/angular/pull/12696
As of angular version 2.3.0-beta.0, the router should be registered on `ng.coreTokens`, so we can take advantage of that and not require the router injection into the root component.

stevenkampen:
resolves #829 

With ng v2.3.0 and angular router v3.3.0 (and above), the router is now resolved automatically (without having to inject it into the root component).

### DIFF
--- a/src/backend/backend.ts
+++ b/src/backend/backend.ts
@@ -322,13 +322,19 @@ export const routersFromRoots = () => {
 };
 
 export const routerTree = (): Array<MainRoute> => {
-  let routes = new Array<MainRoute>();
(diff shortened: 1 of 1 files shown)


rajinder-yadav:
LGTM

stevenkampen:
@rajinder-yadav SQUIRREL???!!?

rajinder-yadav:
:shipit:

PR #919, opened 2017-01-17
Don’t allow spaces in function names

### LINKED ISSUE #699: State not visible when injecting router.
Hey guys,
We use Angular 2.0 + ngrx in our app.

For some reason we can not see the local variables of our component when the component gets the Router injected..

Here is an example:

image

### LINKED ISSUE #903: State of component is not displaying
Augury is a very good tool while developing angular 2 application. But the problem is, it doesn't display state of component. For some components it is showing state, but then some components it doesn't show state. Only dependency is displayed.

stevenkampen:
resolves #699 
resolves #903 

@clbond `functionName` utility produces `bound theFunctionName` for bound functions which broke deserialization. Can you please review this fix?

### DIFF
--- a/src/utils/serialize.ts
+++ b/src/utils/serialize.ts
@@ -125,7 +125,7 @@ function map(operation: Operation, value) {
           /// long as they do not reference closures that are not accessible
           /// in the context they are running in.
           if (typeof value === 'function') {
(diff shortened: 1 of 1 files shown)


rajinder-yadav:
:shipit:

PR #933, opened 2017-01-24
Improve NgModule parsing and protect against missing info in template

### LINKED ISSUE #918: Injector Graph prevent tabs from switching
1. Click on inject graph
2. Click on a component
3. Click on an injector not to make on status info at bottom display something
4. Click on the status info so it has focus
5. Click on a devtools tab

Result: you will not be able to switch between tabs. 

(*) Something the status info fails to show up.

injector-graph

stevenkampen:
resolves #918

### DIFF
--- a/src/backend/backend.ts
+++ b/src/backend/backend.ts
@@ -68,12 +68,16 @@ const messageBuffer = new MessageQueue<Message<any>>();
 /// NOTE(cbond): We collect roots from all applications (mulit-app support)
 let previousTree: MutableTree,
   previousCount: number,
+  initialModules: { [key: string]: any },
   onMouseOver,
   onMouseDown;
 
-const parseInitialModules = (roots: Array<any>) => {
-  if (roots.length) {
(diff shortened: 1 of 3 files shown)


src/backend/backend.ts
```diff
@@ -68,12 +68,16 @@ const messageBuffer = new MessageQueue<Message<any>>();
...
+  initialModules: { [key: string]: any },
   onMouseOver,
   onMouseDown;
 
-const parseInitialModules = (roots: Array<any>) => {
-  if (roots.length) {
-    messageBuffer.enqueue(MessageFactory.ngModules(parseModules(roots[0])));
+const parseInitialModules = () => {
+  const roots = getAllAngularRootElements().map(r => ng.probe(r));
+  if (!initialModules && roots.length) {
```
rajinder-yadav: `if (!initialModules && roots.length > 0)`

roots.length could be a string, so "0" will be true.

rajinder-yadav: Discussed with Steven, he said we should never have to worry about a.length being a string or initialModules being a string, so code as it's written would work correctly.

PR #939, opened 2017-01-27
Chore/update package versions

stevenkampen:
Updates everything to latest versions except basscss.


stevenkampen:
Wow, new mention bot is pushy.

stevenkampen:
@rajinder-yadav Hey, can you sanity check this build as review? It should work and look exactly the same as current dev code. Thanks!

src/frontend/components/injector-tree/injector-tree.ts
```diff
@@ -63,7 +63,7 @@ export class InjectorTree implements OnChanges {
   private unFocusNode = () => {
     this.focusedComponent = -1;
     this.focusedDependency = -1;
-  };
+  }
```
rajinder-yadav: Semicolon at end of the assignment is more correct. In general I prefer we not leave out semicolons.

stevenkampen: @rajinder-yadav I don't mind either way (though I don't use semicolons on class member declarations myself), but since every other instance in this project uses the not-on-class-memeber-decs convention, I would rather have this instance match them (and the current tslint config).

As far as what the convention should actually be, I'm open to changing it, but we should probably defer to some Rangle standard. At any rate it would have to be done in a separate change obviously.

rajinder-yadav: cool

stevenkampen:
@rajinder-yadav Can I get the squirrel here?

stevenkampen:
@rajinder-yadav Can you validate this build (the crx) and then give me a squirrel if it's all good? Thanks!

rajinder-yadav:
Sure I'll take a look at this tomorrow.

PR #944, opened 2017-01-31
Add indent for expandable render-state properties

### LINKED ISSUE #857: Add padding to injector provider tab

stevenkampen:
resolves #857

### DIFF
--- a/src/frontend/components/component-info/component-info.html
+++ b/src/frontend/components/component-info/component-info.html
@@ -31,8 +31,7 @@
           [level]="0"
           [path]="path"
           [state]="instanceProvidersObject"
-          (updateValue)="onUpdateProvider($event)"
-          [indent]="false">
+          (updateValue)="onUpdateProvider($event)">
         </bt-render-state>
       </div>
     </accordion>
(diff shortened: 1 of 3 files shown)


src/frontend/components/component-info/component-info.html
```diff
@@ -51,22 +50,20 @@ <h4 class="primary-color col pt1 pb2 m0">
...
-            <template [ngSwitchCase]="ComponentLoadState.Failed">
-              Failed to load component state
-            </template>
-          </div>
+        <div [ngSwitch]="loadingState">
+          <template [ngSwitchCase]="ComponentLoadState.Received">
+            <bt-render-state
+              [componentMetadata]="componentMetadata"
+              [metadata]="metadata"
+              [level]="0"
```
rajinder-yadav: For a static value you don't need property binding, remove the square brackets, otherwise Angular will try to evaluate the zero.

stevenkampen: @rajinder-yadav I'd rather not change that here since it's not related to this fix at all.

rajinder-yadav: It still incorrect, we should fix this.

stevenkampen: @rajinder-yadav Agreed, it can be improved (remove redundancy), but this code was not added in this PR (the surrounding div was just removed). Let's ship this and make other unrelated changes in another PR (maybe one that corrects all redundant bindings in the app).

rajinder-yadav:
🚢

PR #945, opened 2017-01-31
Fix nested & named outlet route paths

### LINKED ISSUE #914: Route data is not displayed
Using Augury 1.2.8, the Router Tree does not show route's data.

Here is a the repo of a small app that shows the problem: https://github.com/gmarziou/route-data

It defines a route with some data:

~~~typescript
const appRoutes: Routes = [
  {
    path: 'heroes',
    component: HeroListComponent,
    data: { title: 'Heroes List' }
  }
];
~~~

but data is not displayed in route info popup:

2017-01-15_00h05_35

### LINKED ISSUE #917: Router tree, lazy loaded routes show incorrect path in popup
When you hover over a child route node, it shows the subpath as being the root path.

What should be shown is the complete path from the root to the child node.

stevenkampen:
resolves #917 
resolves #914

@rajinder-yadav While addressing this I noticed that we were also not computing the path properly for named outlet routes. What was previously `/parks/park3` should be `/parks/(parkit:park3)`, where `parkit` is the outlet name. This PR fixes that too.

Edit: I've added the fix for #914 here also.

### DIFF
--- a/src/backend/utils/parse-router.ts
+++ b/src/backend/utils/parse-router.ts
@@ -29,7 +29,7 @@ export function parseRoutes(router: any): MainRoute {
   return root;
 }
 
-function assignChildrenToParent(parent, children): [any] {
+function assignChildrenToParent(parentPath, children): [any] {
   return children.map((child) => {
     const childName = childRouteName(child);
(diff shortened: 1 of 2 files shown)


rajinder-yadav:
:shipit:

PR #973, opened 2017-02-13
Dup examples folder deleted.

### LINKED ISSUE #896: Delete kitchen sink project
Kitchen sink project was moved to the augury-examples repo, this issues serves as a reminder to remove dead code from the augury repo.

rajinder-yadav:
resolves #896


stevenkampen:
@rajinder-yadav Awesome. Did we confirm that the kitchen sink app in the augury-examples repo is on par exactly (including linting config, etc)?

rajinder-yadav:
It's a straight copy, so not sure what you mean linting and config? \

The idea is to get rid of kitchen sink app, which Igor and I have discussed. I am just keeping it around for ourselves until as a reference until we're done with the Guides.

However if you feel like it should be kept around, we need to really clean it up, and do a update of all it's parts.

stevenkampen:
@rajinder-yadav If you build the kitchen sink project in `augury-examples` you will see a bunch of linting errors, so it looks like at least the linting config is missing or different. It should be a very simple change, but I think it's worth making it *exactly* the same.

rajinder-yadav:
After discussing with Steven, he would like to keep kitchen sink around for his use, we don't need to make it public through the website that's all.

I also will copy over the Augury linting rule that not also need to be pulling in.

rajinder-yadav:
We get 194 linting errors using Augury's `tslint.conf`, actually kitchen-sink has no linting script so we're good to close this one.

stevenkampen:
@rajinder-yadav What I meant when I mentioned this is that the kitchen-sink application in the `augury-examples` repo produces a large amount of linting warnings, despite reportedly being a straight copy. Before we remove this one, it would be great if the new version behaved in exact parity with this one (it's actually annoying to use for testing with that many linting errors being echoed in the console).

rajinder-yadav:
I've diffed the two folder, other then minor code fix, the only real different I can see is an additional "production" webpack file that was created.

I can't see to figure out why we're seeing more linting error in the augury-example repo. However most seem to be of this type:

```sh
src/components/home.ts[1, 25]: ' should be "
src/components/home.ts[4, 13]: ' should be "
src/containers/kitchen-sink.routes.ts[4, 8]: ' should be "
src/containers/kitchen-sink.routes.ts[6, 31]: ' should be "
src/containers/kitchen-sink.routes.ts[7, 29]: ' should be "
src/containers/kitchen-sink.routes.ts[8, 29]: ' should be "
```

So I generated default lint from from tslint and removed  the rule for, ' should be " and everything now build with no error.

rajinder-yadav:
So let's close this and I will push up the fix to augury-example repo.

rajinder-yadav:
The PR for fix in augury-examples report is here: https://github.com/rangle/augury-examples/pull/49

stevenkampen:
@rajinder-yadav :shipit:

PR #980, opened 2017-02-14
Allow for “root-only” router apps

### LINKED ISSUE #968: Update for angular 2.4.7
Pleace update support for angular 2.4.7

### LINKED ISSUE #970: Augury is very slow or not working
I am using Chrome version 55. 

When I am trying to navigate to different sections like `NgModules` it's very slow or not working at all. 

The second thing is:

alt text

When I am trying to click on the `filters` Input arrow nothing is happening.

### LINKED ISSUE #979: Unable to expand Component State Properties
First of all, thanks for this wonderful development tool 👍 

I am currently experiencing the following issue. Component State Properties which are objects can no longer be expanded/collapsed in the Properties pane. E.g. `_element` in the below screenshot. The issue is independent of the component to be analyzed. This may possibly be related to a recent Chrome update, as I did  not experience the issue before.

I am running: Augury 1.2.11 now on Chrome Version 56.0.2924.87 (64-bit) (Windows 10 machine)

Thanks for looking  into it.

augury-chrome issue

stevenkampen:
resolves #970 
resolves #968 
resolves #979

### DIFF
--- a/src/frontend/components/router-tree/router-tree.ts
+++ b/src/frontend/components/router-tree/router-tree.ts
@@ -52,7 +52,7 @@ export class RouterTree {
     };
 
     const nodes = this.tree(d3.hierarchy(
-      root.children.length === 0 || root.children.length > 1 ? root : root.children[0], d => d.children));
+      (root.children.length === 0 || root.children.length > 1) ? root : root.children[0], d => d.children));
 
     g.selectAll('.link')
       .data(nodes.descendants().slice(1))
--- a/src/frontend/components/router-tree/router-tree.ts
+++ b/src/frontend/components/router-tree/router-tree.ts
@@ -52,7 +52,7 @@ export class RouterTree {
     };
 
     const nodes = this.tree(d3.hierarchy(
-      root.children.length > 1 ? root : root.children[0], d => d.children));
+      root.children.length === 0 || root.children.length > 1 ? root : root.children[0], d => d.children));
 
     g.selectAll('.link')
       .data(nodes.descendants().slice(1))
--- a/src/frontend/components/router-info/router-info.html
+++ b/src/frontend/components/router-info/router-info.html
@@ -38,7 +38,7 @@
           data:
         </span>
         <ul class="list-reset">
-          <li *ngFor="let el of selectedRoute.data">
+          <li *ngFor="let el of selectedRoute._data">
             <div class="pl4">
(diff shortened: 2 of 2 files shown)


stevenkampen:
@rajinder-yadav Can you review this one quickly please? I want to make the new release as soon as possible.

src/frontend/components/router-tree/router-tree.ts
```diff
@@ -52,7 +52,7 @@ export class RouterTree {
     };
 
     const nodes = this.tree(d3.hierarchy(
-      root.children.length > 1 ? root : root.children[0], d => d.children));
+      root.children.length === 0 || root.children.length > 1 ? root : root.children[0], d => d.children));
```
rajinder-yadav: Please add round brackets, no everyone is hip to operator precedence, plus makes the code readable to more people.

I am assuming root.children[0] is a boolean value?

Other than this LGTM.

stevenkampen: I don't understand what you mean. You want brackets around the expression part of the ternary? Is it because it contains an OR? I don't see any issue of precedence (which I would bracket).

BTW, `root.children[0]` is an object (route config), not a boolean.

rajinder-yadav: Just like this, it more for readability for others, not about code correctness :)

```js
root.children.length === 0 || (root.children.length > 1 ? root : root.children[0]), d => d.children));
```

OK so we can have a `boolean` value `(root.children.length === 0)` or an object passed to `d3. hierarchy`, just wanting to make sure as I don't know what this call is doing.

stevenkampen: @rajinder-yadav The ternary expression is always passing an object, but it's either the `root` object, or the first of that object's `children` property (an array).

rajinder-yadav: As discuss two things are happening, one is the short-circuit operation, that will result in a boolean `true` being used, so the ternary operator will never get evaluated. The other is more of a readability edit.

Only on this expression `root.children.length === 0` being `false` will the ternary operator get evaluated.

rajinder-yadav: As further discussed I will need to revisit this short-circuit op, but look like it's working as expected!

:shipit:

Issue #647, opened 2016-09-12
Investigate what "attributes" to show on Component Tree


igor-ka:

clbond:
Hey @sumitarora  do you think I might be able to take this story from you? I would absolutely love to work on this feature & figure out how we can add some directives & additional info to our tree view. But if you are already working on it I'll leave you alone .. just let me know :) thank you

PR #991, opened 2017-02-21
Feature 647 -- attributes on component tree

tyllrb:
Feature 647 -- attributes on component tree

Implementation is pretty straightforward -- extended the node interface to include a property `quickViewAttribute`, which is an array of `Property` objects that the component tree in the front end will display. Once a debug element has been transformed into a node element, the function `quickViewAttribute` will take the node element and take any relevant information based on the node name (currently just `a` elements are considered, but more can easily be added later). 

On the frontend, the `node-attribute` component will take in as an input a nodes `quickViewAttribute`. Initially the component will only display `...` and once the user clicks on the three dots, the array of `Property` objects will be iterated on and displayed (this code is unchanged, it was used previously to display `a` information). 

Thats it, sorry took so long for this to get pushed, took me just a day to wrap my head around the codebase :)

### DIFF
--- a/src/backend/utils/description.ts
+++ b/src/backend/utils/description.ts
@@ -21,27 +21,32 @@ export abstract class Description {
       value: properties,
     });
 
-    switch (provider.name) {
-      case 'RouterOutlet':
-        return p(Description._getRouterOutletDesc(instance));
-      case 'RouterLink':
-        return p(Description._getRouterLinkDesc(instance));
-      case 'NgClass':
-        return p(Description._getClassDesc(instance));
-      case 'NgStyle':
-        return p(Description._getNgClassDesc(instance));
-      case 'NgFormModel':
-        return p(Description._getNgFormModelDesc(instance));
-      case 'NgFormControl':
-        return p(Description._getFormControlDesc(instance));
-      case 'NgControlStatus':
-        return p(Description._getControlStatusDesc(instance));
-      case 'NgModel':
-        return p(Description._getNgModelDesc(instance));
-      case 'NgForm':
-        return p(Description._getNgFormDesc(instance));
+    try {
+      switch (provider.name) {
+        case 'RouterOutlet':
+          return p(Description._getRouterOutletDesc(instance));
+        case 'RouterLink':
+          return p(Description._getRouterLinkDesc(instance));
+        case 'NgClass':
+          return p(Description._getClassDesc(instance));
+        case 'NgStyle':
+          return p(Description._getNgClassDesc(instance));
+        case 'NgFormModel':
+          return p(Description._getNgFormModelDesc(instance));
+        case 'NgFormControl':
+          return p(Description._getFormControlDesc(instance));
+        case 'NgControlStatus':
+          return p(Description._getControlStatusDesc(instance));
+        case 'NgModel':
+          return p(Description._getNgModelDesc(instance));
+        case 'NgForm':
+          return p(Description._getNgFormDesc(instance));
+        default:
+          return p([]);
+      }
+    } catch (e) {
+      return p([]);
     }
-    return p([]);
   }
 
   public static getComponentDescription(debugElement: any): Array<Property> {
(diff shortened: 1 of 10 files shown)


stevenkampen:
@tyllrb I thought the agreement was that we'd not do the ellipsis thing? My thoughts are that if there's value in having it there, it should be visible by default.

src/frontend/components/node-item/node-attributes.ts
```diff
@@ -9,4 +9,11 @@ import {Property} from '../../../backend/utils';
 })
 export class NodeAttributes {
   @Input() private attributes: Array<Property>;
+
+  private show: boolean = false;
+
+  onShow(event) {
+    event.stopPropagation();
```
rajinder-yadav: Does it make sense to eat up the event here. Is it possible that a parent container might want to show something else off this event?

src/frontend/components/node-item/node-attributes.html
```diff
@@ -1,2 +1,9 @@
-<span class="node-item-description" *ngIf="attributes.length > 0">
-  (<template ngFor let-d [ngForOf]="attributes" let-last="last"><p class="node-item-property">{{d.key}}</p><span class="punctuation">=&quot;</span><p class="node-item-value">{{d.value}}</p><span class="punctuation">&quot;</span><span *ngIf="!last" class="punctuation">,&nbsp;</span></template>)</span>
\ No newline at end of file
+<span *ngIf="attributes.length > 0">
+  <span class="node-item-description" [ngClass]="{hide: !show}">
```
rajinder-yadav: Do we need to create a new class for visibility, can you not use " [hidden]" which hide/show an element.

src/frontend/components/node-item/node-attributes.html
```diff
@@ -1,2 +1,9 @@
...
+    (<template ngFor let-d [ngForOf]="attributes" let-last="last"><p class="node-item-property">{{d.key}}</p><span class="punctuation">=&quot;</span><p class="node-item-value">{{d.value}}</p><span class="punctuation">&quot;</span><span *ngIf="!last" class="punctuation">,&nbsp;</span></template>)
```
rajinder-yadav: This line seems too long, can it be better formatted?

src/frontend/components/node-item/node-attributes.html
```diff
@@ -1,2 +1,9 @@
...
+  </span>
+
+  <span (click)="onShow($event)" class="node-item-description" [ngClass]="{hide: show}">
```
rajinder-yadav: Some about visibility as above.

src/tree/transformer.ts
```diff
@@ -210,3 +212,11 @@ const getDependencies = (instance): Array<Dependency> => {
     decorators: parameterDecorators[i] ? parameterDecorators[i].map(d => d.toString()) : [],
   }));
 };
+
+const getQuickViewAttribute = (element): Array<Property> => {
+  if (!element.isComponent) {
+    return quickViewAttribute(element);
+  } else {
```
rajinder-yadav: More of a style, but you don't need and else clause. If you remove it, it is cleaner to read, just simply return an empty array.

tyllrb:
@stevenkampen removed the hide/show functionality for now, like we discussed, we'll see how it plays out with larger apps/component trees and see how best to hide/show the attributes. 

also, before this gets merged, was going to add some failsafe checks to the transformer functions that turns a debug element into a `Node` element, just so if angular debug object changes in the future the entire application won't break due to an exception

rajinder-yadav:
@tyllrb if the Node element changes and the app breaks, that is a good thing. It let's use know we need to fix something. If you hide this (make it fail-safe), the app continues to run crippled and it's going to be hard to find the bug.

stevenkampen:
@tyllrb @rajinder-yadav As per our earlier conversation about this feature, the intention is to make it failsafe.

stevenkampen:
@rajinder-yadav @tyllrb To be clear, this is just the transformer function which derives the "quick view" properties from the node. Not the whole of the transform process.

rajinder-yadav:
@stevenkampen @tyllrb So if the angular debug object changes, how are we going to know this if what it looks like to me is a proposal of eating up this error? Correct me if I am wrong, just want to know how we plan to detect changes outside our control.

src/frontend/components/node-item/node-attributes.html
tyllrb: Only reason this bit is all in one line is because breaking it up with newlines adds unnecessary whitespaces when rendered

src/frontend/components/node-item/node-attributes.ts
tyllrb: This part has actually been removed, there is no click event to listen for anymore

src/tree/transformer.ts
tyllrb: fixed this

src/frontend/components/node-item/node-attributes.html
tyllrb: This part has been removed

tyllrb:
@stevenkampen @rajinder-yadav Added some failsafe checks. In the `quickViewAttribute` function I just added some simple `if` checks to see if certain property names are defined. In the `Description` class, where the angular debug object is used to parse out description information (which is where the `a` quick view attributes come from), I wrapped `getProviderDescription`, and `getComponentDescription` in a `try` statement which will catch the exception when a property name is undefined in the path (ie if `componentInstance` is `undefined` or `null` in the path `debugElement.componentInstance.constructor.name`). 

Right now if the exception is caught those functions will just return `[]`, essentially swallowing the error. Should this error be displayed to the user in any way? If so, this might be outside the scope of this story. I can add failsafe checks for getting the quick view attributes, which I have, but catching and displaying errors/warnings if the entire node transformation fails is of a different concern. 

Also added some changes as per rajinders feedback.

src/backend/utils/description.ts
```diff
@@ -21,64 +21,73 @@ export abstract class Description {
...
-      case 'NgSwitchDefault':
-        return Description._getNgSwitchWhenDesc
+        case 'NgSwitchDefault':
+          return Description._getNgSwitchWhenDesc
           (debugElement.componentInstance);
+        default:
+          return [];
+      }
+    } catch (e) {
+      return [];
```
stevenkampen: @tyllrb This means if one property read fails, that element will have no description properties? Could we easily change this so that only that property is missing in this case?

stevenkampen:
resolves #647

src/backend/utils/description.ts
tyllrb: No, this isn't the case, the description will never have more than one property, hence why each switch case returns immediately once an element type is found and the corresponding property has been read. Only one property read is required for building out the description info (hence the switch statement to determine which property read to perform). If the specific property read fails then the element will have no description properties (empty array), which I think makes sense.

tyllrb: So to address this I created a util function `checkPath`, which will take an object as its first argument and a variable number of arguments to check if the path exists in the object (ex. `checkPath(obj, 'path1', 'path2')` is `true` if obj.path1.path2 exists). Then, in each description function that creates an array of `Property` objects, each property path is checked before including that entry into the array.

stevenkampen:
@tyllrb Excellent. After attending to @rajinder-yadav's suggestions, and a rebase, I think this is a :shipit:

rajinder-yadav:
LGTM :ship:

PR #994, opened 2017-02-23
Support lazy loaded modules in NgModule parsing

### LINKED ISSUE #916: Support lazy loaded modules in NgModule
When a lazy loaded module load into the application, it does not appear in the NgModule tab.

We should refresh the list when a module is loaded.

stevenkampen:
resolves #916

### DIFF
--- a/src/backend/backend.ts
+++ b/src/backend/backend.ts
@@ -1,6 +1,7 @@
 import {Subscription} from 'rxjs/Subscription';
 import {Subject} from 'rxjs/Subject';
 import 'rxjs/add/operator/debounceTime';
+import {compare} from '../utils/patch';
 
 import {
   MutableTree,
@@ -12,7 +13,11 @@ import {
 
 import {onElementFound, onFindElement} from './utils/find-element';
 
-import {parseModules} from './utils/parse-modules';
+import {
+  parseModulesFromRootElement,
+  parseModulesFromRouter,
+  NgModulesRegistry,
+} from './utils/parse-modules';
 
 import {createTreeFromElements} from '../tree/mutable-tree-factory';
 
@@ -33,7 +38,7 @@ import {
 import {send} from './indirect-connection';
 
 import {
-  MainRoute,
+  Route,
   highlight,
   clear as clearHighlights,
   parseRoutes,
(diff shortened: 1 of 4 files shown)


src/backend/backend.ts
```diff
@@ -67,17 +72,33 @@ const messageBuffer = new MessageQueue<Message<any>>();
 
 /// NOTE(cbond): We collect roots from all applications (mulit-app support)
 let previousTree: MutableTree,
+  previousRoutes: Array<Route>,
   previousCount: number,
-  initialModules: { [key: string]: any },
   onMouseOver,
   onMouseDown;
 
+const parsedModulesData: NgModulesRegistry = {
```
stevenkampen: NgModules data object is created initially and added to with each parsing.

src/backend/backend.ts
```diff
@@ -67,17 +72,33 @@ const messageBuffer = new MessageQueue<Message<any>>();
...
+  modules: {},
+  names: [],
+  configs: {},
+  tokenIdMap: {},
+};
+
+const sendNgModulesMessage = () => {
```
stevenkampen: Encapsulated logic to send current NgModule data to the frontend

src/backend/backend.ts
```diff
@@ -106,8 +127,34 @@ const updateComponentTree = (roots: Array<any>) => {
   previousCount = count;
 };
 
-const updateRouterTree = (routes: Array<MainRoute>) => {
-  messageBuffer.enqueue(MessageFactory.routerTree(routes));
+const updateLazyLoadedNgModules = (routers) => {
```
stevenkampen: Called on router change, this function parses any newly loaded NgModules from the routers and updates the frontend.

src/backend/backend.ts
```diff
@@ -106,8 +127,34 @@ const updateComponentTree = (roots: Array<any>) => {
...
+};
+
+const updateRouterTree = () => {
+  const routers: Array<any> = routerTree();
+  const parsedRoutes = routers.map(parseRoutes);
+
+  let routesChanged = !previousRoutes ? true : false;
+
+  if (previousRoutes) {
+    const changes = compare(previousRoutes, parsedRoutes);
```
stevenkampen: Use `compare` to check for changes in the router and only send data to the frontend if there are any.

src/backend/backend.ts
```diff
@@ -227,9 +274,6 @@ const messageHandler = (message: Message<any>) => {
         message.content.path,
         message.content.value));
 
-    case MessageType.RouterTree:
```
stevenkampen: This was never required. Here, the backend responds to messages from the frontend, so this doesn't make sense.

src/backend/utils/parse-modules.ts
```diff
@@ -1,10 +1,18 @@
 import {classDecorators, componentMetadata} from '../../tree/decorators';
 import {functionName} from '../../utils/function-name';
+import {Route} from '../utils/parse-router';
 
 export const AUGURY_TOKEN_ID_METADATA_KEY = '__augury_token_id';
 
 declare const ng;
 
+export interface NgModulesRegistry {
```
stevenkampen: Added interface to describe parsed router data.

src/communication/message-factory.ts
```diff
@@ -6,7 +6,7 @@ import {
 } from './message';
 
 import {
-  MainRoute,
+  Route,
```
stevenkampen: Removed `MainRoute` and replaced usage with `Route`.

stevenkampen:
@rajinder-yadav @tyllrb If you could make a start at reviewing this logic, that would be great. I may still tweak one or two things.

stevenkampen:
@rajinder-yadav I added a small improvement here in another commit. Can you give it a quick review and approve if it's ok?

rajinder-yadav:
:ship:

PR #1006, opened 2017-03-08
Feature/921 router outlet name attribute

### LINKED ISSUE #815: Remove old provider info from mutable tree elements
Currently in the mutable tree we show attributes we parse from the provider info based on a complex Description structure. 

We should remove this info an get it from the provider info in the providers from instanceWithMetadata

### LINKED ISSUE #921: Show name for named router outles
For a named router outlet display its name in the Component Tree.

Picture below has two router outlets, the one at the bottom is named, would be good to display the name inside round brackets like we do for DOM nodes.

screen shot 2017-01-17 at 11 55 27 am

stevenkampen:
resolves #921 
resolves #815


rajinder-yadav:
LGTM

stevenkampen:
@tyllrb You should check out this PR since it kind of wipes your quickViewAttribute stuff. Now the `node.description` property simply holds all attributes that we want to display in this way.

PR #1009, opened 2017-03-09
Fixes error message for non Angular apps.

### LINKED ISSUE #304: Display proper message If not Angular2 application
Several reasons as to when this message can be displayed:
1) This actually NOT and Angular 2 application
2) It is an Angular 2 application but NOT running in DEV mode

The message needs to be generic enough to cover the above scenarios

rajinder-yadav:
resolves #304

### DIFF
--- a/src/frontend/app.ts
+++ b/src/frontend/app.ts
@@ -49,7 +49,7 @@ require('!style!css!postcss!../styles/app.css');
   template: require('./app.html'),
   styles: [require('to-string!./app.css')],
 })
-export class App implements OnInit {
+export class App {
   private Tab = Tab;
   private Theme = Theme;
 
@@ -90,7 +90,7 @@ export class App implements OnInit {
     this.changeDetector.detectChanges();
   }
 
-  ngOnInit() {
+  private ngOnInit() {
     this.subscription = this.connection.subscribe(this.onReceiveMessage.bind(this));
 
     this.connection.reconnect().then(() => this.requestTree());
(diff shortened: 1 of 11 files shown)


src/frontend/app.ts
```diff
@@ -59,7 +59,7 @@ export class App {
   private selectedTab: Tab = Tab.ComponentTree;
   private subscription: Subscription;
   private tree: MutableTree;
-  private error: ApplicationError;
+  private error: ApplicationError = new ApplicationError(ApplicationErrorType.NotNgApp);
```
stevenkampen: @rajinder-yadav Can we change this so this error scenario is explicitly indicated from the backend (or more acurately, the `ng-validate` script)? As it is now we get a flash of this error message while Augury is starting up which I think it a little yucky.

src/frontend/components/render-error/render-error.html
```diff
@@ -14,6 +14,11 @@
...
       </p>
     </div>
+    <div *ngSwitchCase="ApplicationErrorType.NotNgApp">
+      <p>
+        This application is not an Angular application.
+      </p>
+    </div>
     <div *ngSwitchCase="ApplicationErrorType.ProductionMode">
       <p>
         This application is either not an Angular 2 application or is running in production mode
```
stevenkampen: @rajinder-yadav This message here probably needs to be changed since it's now specifically indicating that this application *is* in production mode.

src/frontend/app.ts
rajinder-yadav: I was't able to detect the flashing, but make sense to make the change.

src/frontend/components/render-error/render-error.html
rajinder-yadav: We should be able to detect the Application is not Angular or it's not an Angular app in production mode. I'll consider reviewing once the changes are made to detect this from the backend.

rajinder-yadav:
:shipit:

PR #1030, opened 2017-03-20
Validate origin in message listener

Raz0r:
Hi!

Augury does not validate message origin in event listener which allows attackers to post arbitrary messages to target pages. What is even worse is that data serialization in messages used by the extension is basically evaluation of JavaScript, which makes it possible to gain XSS on any website provided that a user installed the extension.

PoC:

```javascript
targetWindow.postMessage({
	messageSource: 'AUGURY_INSPECTED_APPLICATION', 
	messageType: 1, 
	serialize: 2, 
	content: 'alert(1)'
}, '*')
```


stevenkampen:
@Raz0r Thank you for catching this. We'll do a release right now.

PR #1036, opened 2017-03-22
Display message on browser refresh

### LINKED ISSUE #304: Display proper message If not Angular2 application

rajinder-yadav:
resolves #304


stevenkampen:
@rajinder-yadav :shipit:

PR #1056, opened 2017-03-31
Reselect component on property update

### LINKED ISSUE #1050: Changes to component properties don't work if the value is the same as the original.
The frontend copy of the state is not updated, so any change to the component instance state only are applied if the value is different from the *original* value.

stevenkampen:
resolves #1050 

@cdarken Got it.


rajinder-yadav:
:shipit:

PR #1094, opened 2017-05-04
Feat/analytics2

AbdellaToronto:
src/backend/utils/analytics.ts may be deprecated, if so I'll remove before merging.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
     "@angular/http": "2.4.5",
     "@angular/platform-browser": "2.4.5",
     "@angular/platform-browser-dynamic": "2.4.5",
+    "angulartics2": "^2.0.0",
     "basscss": "^7.0.4",
     "basscss-border-colors": "^2.1.0",
     "basscss-type-scale": "^1.0.5",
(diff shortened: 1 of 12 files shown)


package.json
```diff
@@ -41,6 +41,7 @@
     "@angular/http": "2.4.5",
     "@angular/platform-browser": "2.4.5",
     "@angular/platform-browser-dynamic": "2.4.5",
+    "@angular/router": "3.0.0",
```
sumitarora: Why do we need router?

src/analytics.ts
```diff
@@ -0,0 +1,28 @@
+// needed to refactor the regular GA snippet to be compatible with our tslint
```
sumitarora: we shouldn't need the GA snippet if we are only tracking things manually.

src/backend/utils/analytics.ts
```diff
@@ -0,0 +1,23 @@
...
+  private GA;
+
+  constructor() {
+    if (window.hasOwnProperty('ga')) {
+
+      // todo: figure out what rule to specifically ignore on next line here
+      // instead using type assertion here to get around issue
+      // https://basarat.gitbooks.io/typescript/content/docs/types/type-assertion.html
+      this.GA = (window as any).ga;
+    }
```
sumitarora: Move all the calls https://github.com/rangle/augury/pull/1094/files#diff-0fa1d9c4e14a9a44bcde198728d4435eR22 here you should need `analytics.ts`

sumitarora:
Please rebase too

PR #1102, opened 2017-05-08
fix(router-tree): allow click & drag resize of search pane

### LINKED ISSUE #1062: in router tree, 'search route' cannot verticaly resized
Augury version (required): 1.6.0
Angular version (required): 4.0.0
Date: 
OS: Ubuntu Linux

When I see the 'router tree', I can't to resize - reduce the height of - search routes. And it uses a half of panel height, make the router tree compressend and, sometimes, confuse.

print

riavalon:
Resolves #1062 .


stevenkampen:
:shipit:

Issue #814, opened 2016-11-23
Enhancement: Show available events defined inside a template.


rajinder-yadav:
Currently Augury will only show events that are defined programmatic in the component code using the decorator @Output. However Augury fails to list events defined inside the template.

![screen shot 2016-11-23 at 2 52 24 pm](https://cloud.githubusercontent.com/assets/439226/20576717/0c72c76a-b18d-11e6-89e5-8463bc0b8037.png)

As you can see the button click events is not shown in Augury under properties:

![screen shot 2016-11-23 at 2 52 58 pm](https://cloud.githubusercontent.com/assets/439226/20576728/1a1cd3d8-b18d-11e6-83ea-3cfb5051d82d.png)

It would be nice to make these events available so that can't be triggered for debugging purposes.

PR #1108, opened 2017-05-10
feat(component-info): Display information about template events

riavalon:
Also adds the ability to trigger an event bound to an element.

Addresses #814

### DIFF
--- a/src/frontend/components/component-info/component-info.ts
+++ b/src/frontend/components/component-info/component-info.ts
@@ -61,10 +61,20 @@ export class ComponentInfo {
     return Object.keys(this.state).length > 0;
   }
 
+  private onTriggerTemplateEvent(listener) {
+    this.actions.triggerEvent(this.node, listener);
+  }
+
   private get hasDirectives() {
     return this.node &&
       this.node.directives &&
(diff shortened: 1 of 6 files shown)


src/frontend/components/component-info/component-info.ts
```diff
@@ -59,12 +59,28 @@ export class ComponentInfo {
     return Object.keys(this.state).length > 0;
   }
 
+  private onTriggerTemplateEvent(listener) {
+    chrome.devtools.inspectedWindow.eval(`
```
stevenkampen: @riavalon I *think* we should avoid using `inspectedWindow` here, so that it's consistent with how other Augury -> inspected app communication works. Changing properties, and emitting events all happens by the frontend sending a message to the backend, which then executes the operation. That also makes sure our logic is covered by the error reporting exception handler. Does that sound ok to you?

riavalon: Yeah, makes sense to me, I'll make those changes.

stevenkampen:
@riavalon :shipit:

sumitarora:
@riavalon Please resolve conflicts.

PR #1109, opened 2017-05-11
fix(selected-component): Rename variable to avoid overwrites

### LINKED ISSUE #1044: $a can override a property/function set by the application
Augury version (required): 1.4.0
Angular version (required): 4.0.0
Date: 3/29/2017
OS: Mac

Description of issue:
Augury has this code in backend.ts:
```
const selectedComponentPropertyKey = '$a';

const noSelectedComponentWarningText = 'There is no component selected.';

Object.defineProperty(window, selectedComponentPropertyKey,
  {value: noSelectedComponentWarningText});
```

But that can easily override a $a property/function used by the application. This is especially likely when using Closure Compiler since it will frequently rename a function to $a, and sometimes that ends up being a function on window. I work at Google where I ran into this.

Steps to reproduce:

1. Define window.$a to be a function that is called at runtime by the main application
2. Load the page with Augury installed
3. Get error

riavalon:
Resolves #1044

### DIFF
--- a/src/backend/backend.ts
+++ b/src/backend/backend.ts
@@ -245,7 +245,7 @@ const resubscribe = () => {
 // the Augury UI / frontend is not even open.
 send(MessageFactory.ping()).then(resubscribe);
 
-const selectedComponentPropertyKey = '$a';
+const selectedComponentPropertyKey = '$$el';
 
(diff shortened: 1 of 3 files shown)


igor-ka:
@riavalon LGTM :shipit:

Issue #1025, opened 2017-03-15
UI Issue large blank space under search


rajinder-yadav:
Augury: dev branch
Date: Wed, 15 Mar 2017 23:28:01 GMT
OS: Linux x86_64
Angular version (required): 2.4

Demo test application:
Use Guide routes demo app.

Description of issue:

Large blank area under search.

![big-search-area](https://cloud.githubusercontent.com/assets/439226/23975198/bf5437c6-09b5-11e7-8f3c-af119d2811d6.png)

Steps to reproduce:

1. Open Augury, click on Router Tree

PR #1115, opened 2017-05-16
Fix: Remove empty space under search bar in Router Tree

JinSY:
Addresses #1025


igor-ka:
This PR resolves #1025

PR #1127, opened 2017-05-24
Update prod mode check to reflect changes in angular core

### LINKED ISSUE #1071: "Unhandled Promise rejection: Cannot read property 'injector' of null" CONSOLE ERROR JUST AFTER AUGURY INSTALLATION
Augury version (required):1.6.0
Angular version (required):4.0.1
Date:19/4/2017
OS:Ubuntu

Description of issue:
Augury is not working due to this error in console. This error comes only after I add augury to the list of extensions.

image

stevenkampen:
resolves #1071

### DIFF
--- a/src/utils/ng-validate.ts
+++ b/src/utils/ng-validate.ts
@@ -1,7 +1,14 @@
 import {messageJumpContext, browserSubscribeOnce} from '../communication/message-dispatch';
 import {MessageFactory} from '../communication/message-factory';
 import {MessageType} from '../communication/message-type';
+import {Message} from '../communication/message';
 import {send} from '../backend/indirect-connection';
+
+import {isAngular, isDebugMode} from '../backend/utils/app-check';
+import {ApplicationError, ApplicationErrorType} from '../communication';
+(<any> window).isAngular = isAngular;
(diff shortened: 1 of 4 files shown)


src/utils/ng-validate.ts
```diff
@@ -4,13 +4,15 @@ import {MessageType} from '../communication/message-type';
 import {send} from '../backend/indirect-connection';
 
 declare const getAllAngularTestabilities: Function;
+declare const getAllAngularRootElements: Function;
+declare const ng: any;
 
 let unsubscribe: () => void;
 
 const handler = () => {
   // variable getAllAngularTestabilities will be defined by Angular
```
igor-ka: @stevenkampen seems like you are not using getAllAngularTestabilities anymore. Should this comment be updated? Also, how would that affect previous version support?

stevenkampen: @igor-ka Previous version behaviour should stay the same, but this comment should be updated, and your comment prompted me to think about some other things. Let me improve this PR a little.

stevenkampen:
@igor-ka Can you review this? I'd like to include it in the release which I'll do late this evening.

PR #1155, opened 2017-06-05
Update to Angular v4.1.3

### LINKED ISSUE #1141: Update angular version in Augury frontend

stevenkampen:
resolves #1141


package.json
```diff
@@ -35,14 +35,14 @@
...
     "@angular-redux/store": "^6.4.1",
-    "@angular/common": "2.4.5",
-    "@angular/compiler": "2.4.5",
-    "@angular/core": "2.4.5",
-    "@angular/forms": "2.4.5",
-    "@angular/http": "2.4.5",
-    "@angular/platform-browser": "2.4.5",
-    "@angular/platform-browser-dynamic": "2.4.5",
-    "@angular/router": "3.0.0",
+    "@angular/common": "4.1.3",
```
igor-ka: @stevenkampen let me know if it makes sense to add ^ or not

stevenkampen: @igor-ka I would rather lock everything down, just because I don't see any benefit to the opposite. Any opinion?

JinSY: @stevenkampen I'm in agreement. I would rather keep it locked down

stevenkampen:
@igor-ka Please review again. Had to fix some compilation issues after updating yarn.lock.

PR #1163, opened 2017-06-06
Stop error thrown on no bootstrapped module

### LINKED ISSUE #1162: Sometimes no module can be resolved via the root component
Augury version (required): 1.11.0
Angular version (required): Unknown

stevenkampen:
resolves #1162


sumitarora:
@stevenkampen What happens if the module is not bootstrapped? Do we display any error?

stevenkampen:
@sumitarora No we don't as yet. I'm still not clear on how this can happen. I just want to avoid the error so the user has a chance to see that the NgModules tab is now empty, and hopefully report the error so we can look into it.

Issue #1164, opened 2017-06-06
Extend analytics tracking with additional events


igor-ka:
Refactor code to support and add analytics tracking for the following 2 events:

- emit events
- change of properties

PR #1172, opened 2017-06-07
analytics extension work

JinSY:
implements #1164

### DIFF
--- a/src/frontend/actions/main-actions.ts
+++ b/src/frontend/actions/main-actions.ts
@@ -2,6 +2,7 @@ import {Injectable} from '@angular/core';
 import {dispatch} from '@angular-redux/store';
 import {Tab, StateTab} from '../state';
 import {IAppState} from '../store/model';
+import {Path} from '../../tree';
 
 @Injectable()
 export class MainActions {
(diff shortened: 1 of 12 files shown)


src/frontend/actions/main-actions.ts
```diff
@@ -2,13 +2,17 @@ import {Injectable} from '@angular/core';
 import {dispatch} from '@angular-redux/store';
 import {Tab, StateTab} from '../state';
 import {IAppState} from '../store/model';
+import { Path } from '../../tree';
```
stevenkampen: @JinSY I don't like it either but I think the Augury project is heavily is biased towards omitting the spaces in imports (like the lines above) :). I'd like to keep it as consistent as possible.

src/frontend/components/component-info/component-info.html
```diff
@@ -31,7 +31,8 @@ <h4 class="primary-color col pt1 pb2 m0">
           [level]="0"
           [path]="path"
           [state]="instanceProvidersObject"
-          (updateValue)="onUpdateProvider($event)">
+          (updateValue)="onUpdateProvider($event)"
+          (emitValueChange)="onEmitValueChange($event)">
```
stevenkampen: @JinSY Would this not be better named just `emitValue`?

sumitarora:
@jinSy How can I check the events in Google Analytics ?

src/frontend/components/component-info/component-info.html
JinSY: This is because the `RenderState` component already has an existing `emitValue` identifier but I am making other changes to make this part more consistent with the rest of the project

PR #1202, opened 2017-06-14
feat: allow console reference from elements panel

yannickglt:
This improvement consists in creating a console reference from the DevTools Elements panel as well.

Currently, typing `$$el` in the browser console will only display the component selected in the Augury ComponentsTree.
However, sometimes it may be simpler to get this component reference by clicking on its corresponding element in the native DevTools Elements panel.

Selecting a component from the Augury ComponentsTree still works as before. The current console reference is simply the last selection from both the Augury ComponentsTree and the DevTools Elements panel.


arafeek:
@yannickglt First off, thanks for your contribution to augury! We'll be discussing whether this is a desired feature for an upcoming release. One concern that I have after testing it a bit is that the highlighted element in the `augury` tab and the value of `$$el` can now be out of sync (for example if you select a different element in `elements` and then switch back to `augury`).

yannickglt:
Hi, thanks for your response and for taking the time to discuss about this feature. I gonna check the unexpected behavior you are talking about and suggest a fix (embedded in this pull request)  soon.

Issue #1150, opened 2017-05-31
Feature request: Firefox support


miltador:
It would be awesome to see this extension for Firefox browser. As Firefox supports [WebExtensions](https://developer.mozilla.org/en-US/Add-ons/WebExtensions) API, I believe it won't be difficult to make it working on Firefox. I saw React Dev Tools extension on Firefox, Angular community is now missing this one for Firefox.

captainbrosset:
Firefox now supports a wide range of WebExtension APIs, including DevTools-related APIs. This means that without changing much of the extension's code, it should just work on Firefox.

I tried running it in Firefox and got this:

<img width="1021" alt="capture" src="https://cloud.githubusercontent.com/assets/1152698/26823321/88c16bb4-4aad-11e7-8408-83562e120336.PNG">

Not bad for a first test, at least the panel appears and has some content.
There's an error in the console though: `Augury initialization has failed <unavailable>  content-script.js:5711`.

Looking into it more, the `<unavailable>` bit is actually `Error: Could not establish connection. Receiving end does not exist.`

CC @clarkbw who has been working recently on helping other extensions like react, redux, ember, vuejs become fully Firefox compatible. Maybe he can help here too.

clarkbw:
I'm seeing some intermittent issues here and I don't know Angular well enough to be able to traverse the code.  It seems to drop the connection at times and others it works.  I can pair with someone to get them through the Firefox parts but I'll need help with the rest.

igor-ka:
Hey @captainbrosset and @clarkbw we will be happy with that. We are thinking of a potential refactor to support different shells easier and would appreciate help and/or input on that.

Can we chat over [Slack](https://augury-slack.herokuapp.com/)?

clarkbw:
> Can we chat over Slack?

Sounds good!  I've joined the Slack as @clarkbw 

Lets try to chat soon, ping me when you get a chance @igor-ka

digitarald:
Hi @igor-ka , I am taking over this effort from @clarkbw , working with @captainbrosset . To kick things off I joined Slack and will ping you.

adityamenon:
Hello all, I do not see anything on the README or an Addon on Firefox yet. Which GitHub issue are we tracking the progress for this on?

karlos1337:
Something about this?

Invis1ble:
bump

andrewthauer:
[...]

PR #1278, opened 2018-03-27
Firefox Text & Documentation Updates

andrewthauer:
Text & documentation updates for pending Firefox add-on support #1150.


andrewthauer:
This is just a start focusing on Firefox support. At some point all the material should be thoroughly reviewed for relevancy and potential updates.

Issue #1289, opened 2018-04-18
Augury breaks Firefox XML viewer


andreicristianpetcu:
-- please install the [Augury Canary Build](https://chrome.google.com/webstore/detail/augury-canary-build/leechohfifidanmkioncamdcibhmmcji?hl=en) Chrome extension, as this will be used to test any new features or bug fixes.

Augury version 1.16.0:
Angular version any:
Date: 18th of April 2018
OS: GNU/Linux

Description of issue:
-- When I install Augury I cannot see XML files highlighted in Firefox.

Steps to reproduce:

1. Visit any XML page with Firefox without Augury installed for example http://search.maven.org/remotecontent?filepath=com/jenmaarai/sidekick/2.5/sidekick-2.5.pom and you will see a pretty XML.
![screenshot from 2018-04-18 18-48-46](https://user-images.githubusercontent.com/922368/38947760-0c7746ac-433e-11e8-885f-e686a8321845.png)

2. Install Augury and visit the same page and you will get no XML formatting.
![screenshot from 2018-04-18 18-48-52](https://user-images.githubusercontent.com/922368/38947783-17e93c0c-433e-11e8-89a8-c41bce9711f9.png)

santiago-elustondo:
Thanks @andreicristianpetcu. Please see PR comments.

santiago-elustondo:
merged

PR #1290, opened 2018-04-18
Fix firefox xml viewer

andreicristianpetcu:
Fix for this issue https://github.com/rangle/augury/issues/1289


src/content-script.ts
```diff
@@ -37,56 +37,60 @@ const injectScript = (path: string) => {
 
 export const injectSettings = (options: SimpleOptions) => {
   inject(script => {
-    const serialized = JSON.stringify(options);
+    if (document instanceof HTMLDocument) {
```
santiago-elustondo: Why do we need to check `instanceof` here as well? Seems redundant.

src/content-script.ts
```diff
@@ -37,56 +37,60 @@ const injectScript = (path: string) => {
...
-      // because that allows us to send the correct tree immediately upon
-      // startup and send it to the message queue, allowing Augury to render
-      // instantly as soon as the application is loaded. Without this bit
-      // of code we would have to wait for the frontend to start and load its
-      // options and then request the tree, which would add a lot of latency
-      // to startup.
-      injectSettings(options);
-
-      injectScript('build/backend.js');
+if (document instanceof HTMLDocument) {
```
santiago-elustondo: This seems reasonable.

andreicristianpetcu: Thank you @santiago-elustondo for reviewing my changes. I reverted the other change since it was indeed not needed.

PR #1296, opened 2018-04-30
Build conditionals

santiago-elustondo:


src/build.ts
```diff
@@ -0,0 +1,23 @@
...
+  enableGoogleTagManager: Boolean(parse(GTM)),
+  enableSentry: Boolean(parse(SENTRY)),
+  PROD_MODE: Boolean(parse(PROD_MODE)),
+};
+
+function parse(stringified) {
+  return JSON.parse(stringified || null);
+}
+
+export default BUILD;
```
andrewthauer: I'm not a huge fan of default exports. Can we make this a named export. Also can we call it `buildConfig` or something.

src/build.ts
```diff
@@ -0,0 +1,23 @@
+
+// expected injected enviornment variables
+declare const GTM: any;
+declare const SENTRY: any;
+declare const PROD_MODE: any;
+
+interface BuildConfig {
+  enableGoogleTagManager: boolean;
+  enableSentry: boolean;
+  PROD_MODE: boolean;
```
andrewthauer: Can we make this naming consistent (e.g. `prodMode`)?

src/build.ts
```diff
@@ -0,0 +1,23 @@
+
+// expected injected enviornment variables
+declare const GTM: any;
+declare const SENTRY: any;
+declare const PROD_MODE: any;
+
+interface BuildConfig {
+  enableGoogleTagManager: boolean;
```
andrewthauer: Nit picking here, but I personally think starting with `enableFoo` suggests it's an action when in fact it's intention is to be a flag. I think `googleTagManagerEnabled` would be more clear. Or you could create a nested key likes so:

```ts
{
  googleTagManager: {
    enabled: boolean;
  }
}
```

webpack.config.js
```diff
@@ -138,27 +130,8 @@ module.exports = {
...
-}
-
-function manifestFiles() {
-  return [
-    // base manifest file
-    'manifest/base.manifest.json',
-    // each build can extend the base manifest with a file of this form
-    `manifest/${targetBuild().toLowerCase()}.manifest.json`,
-  ];
+function stringifyValues(obj) {
```
andrewthauer: You could perhaps move this to the `build.config.ts` file to keep the webpack file a little cleaner.

PR #1333, opened 2018-08-16
Fix Firefox Manifest CSP Issue

andrewthauer:
The current Firefox manifest was not allowing evals (which are required) for the serialization process. This was causing state to not show in the inspector.


andrewthauer:
Not specifically, but it will now allow using node 10+. Figured it would be good to allow it since 10 will be the active LTS in October.

Issue #1340, opened 2018-09-10
Augury not showing state for many components


tkstang:
Augury version (required): Canary build 1.19.3
Angular version (required): 6.0.7
Date: 9/10/18
OS: Mac OS High Sierra

I know this has been an on going issue that seems to have been a problem for many people, all previously opened issues were marked as resolved so I'm opening a new one.  I'm using the most current canary build and many of my components still load and then indicate "no state to show" in the properties tab.  Unfortunately I can't share a repository as the issue is with a private repo at my place of work.  The application is rather complex, it seems that simpler smaller components will show state but most larger components do not.  I have previously seen in other issues where there were mentions of things like components with router as a dependency working, components loaded with resolvers, etc.  All of these issues are currently marked as resolved, however, and I am still experiencing the issue.

santiago-elustondo:
Hey, @tkstang 
Thank you for reporting.

Does the state tab show "no state to show", "loading state..." or any other message?

Could you open a second inspector window, and tell me if you see any errors in the console? (hit option+command+i while looking at Augury's devtools panel)

tkstang:
It shows "loading state..." for a few seconds and then switches to "no state to show."  I'm not sure if I'm not quite understanding you correctly but hitting option+cmd+i while looking at augury is just going to close the inspector window.  If you mean errors with the application there are no errors aside from a server error from a network request our back end that is down.  Nothing with the application itself.  If there is some way I'm missing to actually see any errors with augury itself perhaps I'm not quite following you.

tkstang:
I've attempted to do this while looking at Augury and this is simply closing the original inspector window.

santiago-elustondo:
Ok, so we have another edge case affecting the serializer. We need to figure out why there is a syntax error in the generated serialized message.

MattMattV:
On Manjaro with :
  - Angular version 6.1.7
  - Augury version 1.19.1 AND 1.19.3
  - Chrome 70.0.3534.4 version 

I face a very similar issue, the message "No state to show" appear on some components but with a different error in the inspector :

```
Error: Code evaluation failed
    at Array.<anonymous> (chrome-extension://elgalmkoelokbchhkhacckoklkejnhcd/build/backend.js:31)
    at i (chrome-extension://elgalmkoelokbchhkhacckoklkejnhcd/build/backend.js:31)
    at Object.e.serialize (chrome-extension://elgalmkoelokbchhkhacckoklkejnhcd/build/backend.js:31)
[...]

Issue #1348, opened 2018-11-04
State failing to load for `router-outlet` & native elements


andrewthauer:
Some components (e.g. `router-outlet`) & native elements (e.g. `a`, etc.) do not show state. These are being serialized correctly, but the `metadata/instanceWithMetadata` method does not recognize these as components and thus the `InstanceWithMetadata` structure is not loaded correctly.

In order to fix this, the `instanceWithMetadata` function will need to accommodate non components and build an appropriate InstanceWithMetadata structure consistent with components so they show correctly in the state inspector panel.

PR #1349, opened 2018-11-04
Fix Serializer Implementation

andrewthauer:
The current serialize/deserialize implementation has been problematic and the cause of many bugs. Swapped out the custom `serializer` to use [flatted](https://github.com/WebReflection/flatted) which handles circular references + seems to be much more resilient to a broad range of object structures.

I've tested on a few different apps and seems to work on almost all components correctly. With the exception of `router-outlet` & native elements (see #1348).


andrewthauer:
It seems to work a heck of a lot better imo. I've tested with a few apps. It also fixes #1340 for angular/fire. Changing state and emitting in the state panel also seem to work. Haven't noticed any regressions. Was going to push a canary build and see.

stevenkampen:
@andrewthauer 👍 good one!

PR #1352, opened 2018-11-19
Add HMR support

stevenkampen:
@andrewthauer This adds HMR support. It works in my local HMR configured app and on stackblitz.com (e.g. https://angular-9b55kc.stackblitz.io/).

### DIFF
--- a/src/backend/backend.ts
+++ b/src/backend/backend.ts
@@ -61,6 +61,8 @@ import {SimpleOptions} from '../options';
 
 import { MessagePipeBackend } from 'feature-modules/.lib';
 import { highlighter } from 'feature-modules/highlighter/backend/index';
+import { ApplicationRef, NgModuleRef } from '@angular/core';
+import { Observable } from 'rxjs';
 
 declare const ng;
 declare const getAllAngularRootElements: () => Element[];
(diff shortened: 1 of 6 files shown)


src/utils/serialize.ts
```diff
@@ -18,5 +18,5 @@ export const serialize = value => {
 
 /// Deserialize a function string and invoke the resulting object recreator.
 export const deserialize = value => {
-  return parse(value);
+  return value === 'null' || value === null ? null : parse(value);
```
andrewthauer: Curious if this was related to HMR or does it potentially resolve some of the reported regressions?

src/backend/backend.ts
```diff
@@ -214,31 +205,20 @@ const updateRouterTree = (): Promise<void> => {
...
-    subject.debounceTime(0).subscribe(() => {
-      Promise.all([
-        updateComponentTree(getAllAngularRootElements().map(r => ng.probe(r))),
-        updateRouterTree()
-      ]).then(() => onUpdateNotifier.next());
-    }));
-
-  // initial load
-  subject.next(void 0);
+const listenForSomeTimeAndMaybeResubscribe = (timeMs: number) => {
```
andrewthauer: Any chance something like this approach might work in `augury-labs` lazy loaded module support?

src/utils/serialize.ts
stevenkampen: @andrewthauer This just avoids passing the `"null"` string into `parse` which throws an error. This only occurs (the `value` of `"null"`) when some error has happened in the backend (I think  with the serialization).

andrewthauer:
@stevenkampen - Did you want to do a canary release of this fix and see which issues it fixes?

src/backend/backend.ts
stevenkampen: @andrewthauer Possibly, but it's all still a little hazy for now.

Issue #1344, opened 2018-09-14
Component change detection is incorrectly shown as default


mjolka:
Augury: 1.19.1
Date: Fri, 14 Sep 2018 02:18:14 GMT
OS: Win32

Demo test application:
https://change-detection-strategy-onpush.stackblitz.io/

Description of issue:
Components that use ChangeDetectionStrategy.OnPush are incorrectly shown to be using ChangeDetectionStrategy.Default

![change detection](https://user-images.githubusercontent.com/6613722/45526077-14f39100-b819-11e8-9405-d447103319d6.png)

Angular version (required): 6.0.0

Steps to reproduce:

1. Create a component using ChangeDetectionStrategy.OnPush
2. Inspect the component's change detection strategy using Augury

Additional details:

Possibly related to https://github.com/angular/angular/issues/21627

mhagnumdw:
When will this fix be released?

mboughaba:
It is working properly now thanks.
however component state is empty so one needs to inspect it in console using `$$el`

stevenkampen:
@mboughaba For issues with the component state not being serialised and displayed properly I'll need direct instructions to reproduce, since those issues are particular to *your* state. So, if you can work out what it is in your component that is triggering the error scenario, and recreate it in a stackblitz app, I'll look at it immediately. Otherwise there's nothing I can do.

mhagnumdw:
> @mhagnumdw This should be released now in version 1.22.0. I've just published, so before you verify, make sure your extension has actually updated to 1.22.0.

Now it works!

But, now I notice a new problem: when removing or adding a component in the code, the view (browser) is updated, but the "Component Tree" does not. If I remove a component from the code, it continues in the "Component Tree" and clicking it causes the following error:

![image](https://user-images.githubusercontent.com/17688846/50489277-8086f600-09e5-11e9-99df-04ad19d9d3b6.png)

PR #1353, opened 2018-11-22
#1344 Add new component metadata resolution for ng v5+

stevenkampen:
@andrewthauer This fixes: https://github.com/rangle/augury/issues/1344

Starting around angular 5 the component metadata was handled slightly differently. This *should* be fully backwards compatible, but I can't say I have a bunch of older angular apps to test with. How have you been managing that? Can stackblitz be used for older ng versions?


andrewthauer:
@stevenkampen - StackBlitz has been around since at least Angular 5 (possibly 4), so I suspect you can just change the dependency versions for testing.

danmana:
Is it possible to get this merged?

I can confirm it works with Angular `4.1.3` and `6.1.7`

Issue #1320, opened 2018-06-22
slow and often leads to browser crashes


xunjianxiang:
-- please install the [Augury Canary Build](https://chrome.google.com/webstore/detail/augury-canary-build/leechohfifidanmkioncamdcibhmmcji?hl=en) Chrome extension, as this will be used to test any new features or bug fixes.

Augury version (required): 6.0.5
Angular version (required): 1.19.1
Date: 2018-06-22
OS: macOS Hign Sierra 10.13.4
UI Framework: ng-zorro-antd

-- Please make sure you're using the latest version of Augury before reporting an issue.

Demo test application:
-- Git repository for demo app showing the issue (optional but very helpful for difficult issues).
-- If a code snippet will completely show the issue, please include it.

Description of issue:
-- Include (clipped) screenshot images if possible.

![image](https://user-images.githubusercontent.com/12422146/41774425-467d5390-7652-11e8-97f7-eb3215254785.png)

![image](https://user-images.githubusercontent.com/12422146/41774450-59decad6-7652-11e8-950f-cd2d2cf20301.png)

Steps to reproduce:

1. any way

Additional details:

None

andrewthauer:
@xunjianxiang - Can you provide some additional details that might help reproduce this? For instance:

- Does this occur right away or after your using augury? if so how long and what are you doing when it crashes?
- How big is the application?
- Are you running a lot of other chrome tabs, etc.
- Do you have a sample app you could share?

andrewthauer:
Could be related to #1317?

Benno007:
I've been having similar issues with our angular 5 application in development mode.

Memory footprint of Augury in chrome task manager shows it starts off at about 40mb then jumps to 300-400mb when inspecting some components properties. At this point, if I interact with my application, the chrome tab process spikes to 200+ % CPU and becomes unresponsive, ultimately I have to kill the tab to be able to use the app again. This happens every time I use Augury to inspect a component that has a fair amount of state with observable properties and also with services. The CPU spike only seems to happen if I'm trying to interact with my application while its loading the state of the component

zaalbarxx:
I also have some problems regardin Augury recently. It works really sluggish, after inspecting component I have to wait for 30+ seconds to actually get the component data, the dev tools freeze for this waiting time and sometimes the whole page crashes.

webia1:
same here,..

khaledalyawad:
same here!

stevenkampen:
[...]

PR #1417, opened 2019-11-15
Sanity for AppRef.isStable observable

feeloor:
In larger applications, we sometimes encounter the `appRef.isStable` observable to fire constantly which is causing us to constantly send pings and updates, this in turn is making larger applications almost unusable. 
To prevent this I've added a sanity check to make sure that we only emit at the most twice per second.

Related issues: #1320 possibly #1402

### DIFF
--- a/src/backend/backend.ts
+++ b/src/backend/backend.ts
@@ -223,13 +223,18 @@ const resubscribe = () => {
             const roots = collectRoots();
             if (roots.length) {
               let sanity;
+              // Adding sanity threshold to make sure
+              // larger app's doesn't get flooded
               const sanityThreshold = 0.5 * 1000; // 0.5 seconds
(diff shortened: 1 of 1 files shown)


src/backend/backend.ts
```diff
@@ -229,26 +217,34 @@ const resubscribe = () => {
...
-          }
-        });
-      })
+      Promise.resolve()
+        .then(() => {
+          runAndHandleUncaughtExceptions(() => {
+            const roots = collectRoots();
+            if (roots.length) {
+              let sanity;
+              const sanityThreshold = 0.5 * 1000; // 0.5 seconds
```
sumitarora: @feeloor Please add a comment explaining what is going on. Rest looks good to me.

feeloor: @sumitarora I've now added a few comments. 👍

Issue #1365, opened 2019-01-30
Component Tree does not populate


zakmac:
**Date** 01/30/2019

| Software | Version |
| -------- | ------- |
| Augury | 1.22.0 |
| Angular | 6.1.2 |
| Windows | 10 |
| Chrome | 71.0.3578.98 |
| FireFox | 64.0.2 |

**Demo test application:** n/a

**Description of issue:**
* Augury developer tool opens, but displays a blank component tree
* Screenshot shows Angular 6.1.2 application, the issue persists even after updating to Angular 7.2.2
![image](https://user-images.githubusercontent.com/2379809/52004698-9bcda100-247c-11e9-9ee6-1e626282cb23.png)
* Other functionality such as the Router Tree/ngModules function as expected
![image](https://user-images.githubusercontent.com/2379809/52004818-f404a300-247c-11e9-8825-9cc64dd642a9.png)
* Augury functions as expected on other sites running an Angular application
  * For example, [Tutorial of Heroes](https://stackblitz.com/angular/endakevqpjj) loads the Augury Component Tree tab without issue
![image](https://user-images.githubusercontent.com/2379809/52005201-fa474f00-247d-11e9-9c71-b7eb978e9b9a.png)

**Steps to reproduce:**
Can be reproduced 100% of the time in our Angular application by simply loading the application and opening Augury from DevTools. Unfortunately (for us) Augury works just fine on other Angular applications such as the Tutorial of Heroes demo app.

**Additional details:**
* Augury has worked fine in the past on this very same application. I am not sure if there are limits to the number of components shown which we have violated by having a larger application? There is no such indication in the UI
* Using the "search components" input under the Component Tree window to search for a known component on-page does not return results

andrewthauer:
@zakmac - Are there any errors? If you right-click / inspect the devtool tab, you might seen console errors. Those could prove useful for troubleshooting.

taylanken:
I'm also having issues with the component tree not populating.


Software | Version
-- | --
Augury | 1.23.0 (canary)
Angular | 7.2.15
Windows | 10
Chrome | 78.0.3904.70

When I click on the Augury-tab in Chrome DevTools, I get 2 errors on my console:

```
Unhandled Promise rejection: Cannot set property root of [object Object] which has only a getter ; Zone: <root> ; Task: Promise.then ; Value: TypeError: Cannot set property root of [object Object] which has only a getter
    at o (backend.js:31)
    at o (backend.js:31)
[...]

Issue #1362, opened 2019-01-13
Component tree not updated


mhagnumdw:
**Versions:**
Augury version (required): after version 1.21.0
Angular version (required): 6.1.8
Date: 01/13/2019 (MM/dd/yyyy)
OS: Windows 7 Pro 64 bits / Fedora 29

**Version 1.21.0 works fine.
Currently affected versions: 1.22.0 and Canary**

**Description of issue:**
Component tree not updated

**Steps to reproduce:**
1. when removing or adding a component in the code or navigating to another component
2. the view (browser) is updated, but the "Component Tree" does not
3. the old/previous component continues in the "Component Tree" and clicking it causes the following error:

![image](https://user-images.githubusercontent.com/17688846/51085477-19757b00-1721-11e9-9df6-4faaf7719b2a.png)

askarby:
I'm having the same issue, rendering Augury pretty much useless at the moment.

alindberg:
I have the same issue.  In addition, a reload/refresh of the page doesn't update the display either.

The Component Tree will refresh only if you move to 'Router Tree'  or 'NgModules' then back.  Not very useful.

```
Augury 1.22.0
Chrome 71.0.3578.98
Angular CLI: 7.1.4
Node: 8.11.2
OS: win32 x64
Angular: 7.1.4

```

s001dxp:
Same issue. However the only way to get augury to update is by exiting DevTools window and reopening.

sm-g:
workaround: change theme (dark/ligth) mode

Issue #856, opened 2016-12-05
Provide a way to refresh angury when using webpack Hot Module Replacement


aitboudad:
I get the following message:
```console
Cannot find element associated with node 0 0 0
```

mobinni:
@aitboudad My initial thought is this is because HMR hotpatches your code and whatever that does, breaks augury from correctly parsing the Angular root element. We will need to research this in order to confirm it. A potential fix for you could be to introduce polling with webpack, for a full page-refresh.

aitboudad:
check https://github.com/jschwarty/angularcli-hmr-example

```
npm run serve:hmr
```

mobinni:
@aitboudad Thank you, we will try to get to this as soon as possible

aitboudad:
when updating the title of `AppComponent` the state component disappear

![augry](https://cloud.githubusercontent.com/assets/1753742/20898168/45c9010c-bb1d-11e6-99c3-2fdcdef1616a.gif)

I have a difference behavior on my project but I couldn't reproduce it on the example app.

aitboudad:
I solved my issue with hacky way :)
```ts
module.hot.dispose(() => {
    ....
    ngModule.injector.get(TestabilityRegistry)._applications.clear();
    ngModule.injector.get(NgZone)._onStable.emit();
   ...
});
```

PR #1436, opened 2019-12-10
Add refresh button

feeloor:
Adding refresh button to manually refresh Augury.

Screen Shot 2019-12-10 at 1 37 33 PM

will help with #1362, #856, possibly helping #1365


yallen011:
Can we make the refresh button a refresh icon? i like the icon idea. You can get an icon from Material icons and give it any color you want. it could even be purple if you want :) But other then that. LGTM

PR #1452, opened 2020-01-28
Fix language of code block

munierujp:
Comment is invalid in JSON.
So, it should be use JSON5 instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,7 +146,7 @@ export default class KitchenSink {
 
 In order for Angular to expose the debug information for AoT applications, you will have to explicitly set the debug flag to `true` in your project's `tsconfig.json` as such:
 
-```json
+```json5
 "angularCompilerOptions": {
   /* ... */
   "debug": true


AleksanderBodurri:
This is inaccurate. The typescript team does not use JSON5.

See https://github.com/Microsoft/TypeScript/issues/12242

munierujp:
jsonc? 

```jsonc
"angularCompilerOptions": {
  /* ... */
  "debug": true
}
```

https://code.visualstudio.com/docs/languages/json#_json-with-comments

AleksanderBodurri:
@munierujp 

According to the link you sent jsonc is the schema used for VS code configuration files, nothing to do with typescript.

Further down `tsconfig.json` is mentioned and it says that the schema to parse it is defined by the VS code extension for typescript.

Regardless the comment in question is meant more as an indication that you _could_ have more configuration above, not that you must have a comment there.

